PR #881, opened 2018-12-10
Fix (Tabs): 1128 Accessibility Missing form label

gepifanio:
[#1128](https://github.com/unimelb/find-a-course/issues/1128)

### DIFF
--- a/components/tabs/Tab.vue
+++ b/components/tabs/Tab.vue
@@ -1,8 +1,7 @@
 <template>
   <transition
     name="fade"
-    mode="out-in"
-  >
+    mode="out-in">
     <div
       v-show="isActive"
       :id="`${namespace}-panel-${index + 1}`"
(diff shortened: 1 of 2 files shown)


gepifanio:
Deploy preview for *pattern-lib-unimelb* ready!

Built with commit f90ae61fe4eca6544fe176045b2359efc6d760dd

https://deploy-preview-881--pattern-lib-unimelb.netlify.com

components/tabs/Tab.vue
```diff
@@ -1,12 +1,12 @@
...
   <transition
     name="fade"
-    mode="out-in"
-  >
+    mode="out-in">
     <div
       v-show="isActive"
       :id="`${namespace}-panel-${index + 1}`"
       :aria-labelledby="`${namespace}-${index + 1}`"
+      :aria-hidden="! isActive"
```
meligatt: @gepifanio In general the usage of `!` before a boolean is without keeping a space in between. lke this: `!isActive`

meligatt: @gepifanio By the way: Why do you need to set aria-hidden: true/false in this div?

PR #912, opened 2019-02-07
feat(component): In Page Navigation component

gepifanio:
Closes tickets:
[#21](https://app.zenhub.com/workspaces/ffam-prototype-5c615588801d5a103031639c/issues/unimelb/ffam-prototype/21)
#911 

Feature In-Page Navigation Component


gepifanio:
Deploy preview for *pattern-lib-unimelb* ready!

Built with commit 07bcfc24551bda52c57de8f2452af4f11fac188d

https://deploy-preview-912--pattern-lib-unimelb.netlify.com

components/page/navigation/InPageNavigation.vue
```diff
@@ -0,0 +1,106 @@
...
+    },
+    headingLevel: {
+      type: String,
+      required: true,
+    },
+  },
+  computed: {
+    getInPageData() {
+      const pageNav = [];
+      document.querySelectorAll(`${this.headingLevel}`).forEach((element) => {
```
meligatt: In terms of component good practices we don't want to create dependencies inside a component from the outer world. It is recommendable to follow the `top down` pattern which you send props to a components to render something, in this case we want a prop to receive a list of the headings to render in the component. 

More info: 
`https://hackernoon.com/10-react-mini-patterns-c1da92f068c5`
and check the section: `#1 Sending data down and up`

In terms of responsibilities, The CMS should know what data to pass to the component, so this one just have the responsibility of "rendering the data". That way we avoid rendering the wrong information.

meligatt: Vue top down approach: https://vuejs.org/v2/guide/components-props.html#One-Way-Data-Flow

PR #939, opened 2019-02-26
feat(component): mega menu with top items

gepifanio:
[45](https://app.zenhub.com/workspaces/ffam-prototype-5c615588801d5a103031639c/issues/unimelb/ffam-prototype/45)

## Design: 

https://zeroheight.com/17lwnkf17/p/59119/i/800676


gepifanio:
Comment: Deploy Started

Building with commit 7aaccfb96e6a84ea2031c66150a33856b9ad95b9

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5c74d0dc627b700008cb0cc6

meligatt:
@gepifanio One thing that I would like to know is "fix the css" what does that mean?
could you provide more specifics hints for me to help 👍 
remember that we have limited time in genreal, so everytime you need some help on specifics it's good to help us give details of what is the problem

TimMansell:
Was there any reason to create a new component <mega-menu-top> instead of adding the features to <mega-menu>?  It seems like a lot of duplicated code.

I'd recommend combining into <mega-menu> and hide/show elements depending if any values are passed.

Create a <mega-menu-secondary-nav> child component and move the secondary menu markup to that.  The mega-menu component is pretty complex so, would componatise it a bit better.

gepifanio:
> @gepifanio Was there any reason to create a new component instead of adding the features to ? It seems like a lot of duplicated code.
> 
> I'd recommend combining into and hide/show elements depending if any values are passed.
> 
> Create a child component and move the secondary menu markup to that. The mega-menu component is pretty complex so, would componatise it a bit better.

@TimMansell  I started using the existing one, but after the high fidelity designs were ready, because the alignment and some details are slightly different from the existing menu, I thought would be better to split in two components, however if you think is better keeping using the existing one, and just play with the `v-if` we go with that not a problem.

just as a reference follow the existing one and the ffam design:

## existing
image

## ffam 

image

TimMansell:
@gepifanio, no problem.  Try it in the existing component and see how that goes.

gepifanio:
> @gepifanio, no problem. Try it in the existing component and see how that goes.

The first commit, was using the existing MegaMenu to build the ffam menu, and I found some limitations to use the same one to match the design, I am not sure if we can build a good component in that way because the design is kinda different... should we ask Jared to see if it is all good to the new menu be more like the existing one instead of the one that he designed?
https://github.com/unimelb/pattern-lib/pull/939/commits/9c13771fe5ad21aa67b02039537af5ece648dc82

gepifanio:
@TimMansell I follow your recommendation, and create the **sub-menu** using the existing component, could you have a look and see if we go that way?

components/page/megamenu/MegaMenu.vue
```diff
@@ -30,10 +31,21 @@
             role="button"
             aria-label="Close"
             class="menu__back-btn"
-            @click="dismissMobileMenu">Close</div>
+            @click="dismissMobileMenu"
+          >Close</div>
           <PageSearchForm
             v-if="isMobileOpen"
             aria-hidden="true"/>
+          <div
```
meligatt: @gepifanio maybe this ui can be a new child component? so the logic+template is scoped and we don't overload the parent (megamenu) with extra responsibilities..

components/page/megamenu/MegaMenu.vue
```diff
@@ -127,8 +142,9 @@
             role="presentation"
             focusable="false"
             aria-labelledby="icon-menu"
-            viewBox="10 10 26 28">
-            <path d="M6 36h36v-4H6v4zm0-10h36v-4H6v4zm0-14v4h36v-4H6z" />
+            viewBox="10 10 26 28"
```
meligatt: @gepifanio I think Raw SVG markup could be in a component...maybe...? seeing `<path ...>` (svg markup) in the component seems unnecessary..

components/page/megamenu/MegaMenu.vue
```diff
@@ -160,6 +176,10 @@ export default {
       type: [String, Boolean],
       default: false,
     },
+    topItems: {
```
meligatt: @gepifanio `topitems`  as a name I think is vague mayeb you can use soemthing like `topNavigationItems` ?

components/page/megamenu/MegaMenu.vue
```diff
@@ -160,6 +176,10 @@ export default {
...
+      type: Array,
+      default: () => [],
```
meligatt: @gepifanio why you need an arrow function to return the default empty array?

components/page/megamenu/index.css
```diff
@@ -4,6 +4,28 @@
...
+.menu-top {
+  display: flex;
+  position: absolute;
+  z-index: 2;
+  right: 98px;
+  margin-top: 14px;
+
+  &__item {
+    margin-left: 20px;
+    color: #adcdf0;
```
meligatt: @gepifanio don't forget to check if this color `adcdf0 ` already is available via css variables.

components/page/megamenu/index.css
```diff
@@ -4,6 +4,28 @@
   --tr-transform: transform .3s ease;
 }
 
+.menu-top {
+  display: flex;
+  position: absolute;
```
meligatt: @gepifanio 
A few things:

- This can be done simpler without mixing position absolute and flexbox check this PR with example: https://github.com/unimelb/pattern-lib/pull/941 
- I don't recommend to use z-index to fix the issue of the top menu rendering behind the other menu.

components/page/megamenu/MegaMenu.vue
gepifanio: When the type is specified and set to Array or Object literal value is not valid, so like `default: ''` is not a valid value.

meligatt: Ok, be mindful when returning an empty "object" using arrow function instead of normal function.

ES6 syntax for objects is `() => ({})`, as `() => {}` treats the curly brackets as a block statement that returns undefined.
```
props: {
  arr: {
    type: Array,
    default: () => ({})
  }
}
```
reference: https://github.com/vuejs/vue/issues/1032#issuecomment-462263430

meligatt: ARRAY:
es6
`default: () => []`
es5
default: function(){ return [] }

Object:
es6
`default: () => ({})`

es5
default: function(){ return {} }

This will return "undefined" instead of empty object.
es6
`default: () => {}`

components/page/megamenu/_mega-menu.css
```diff
@@ -4,6 +4,28 @@
   --tr-transform: transform .3s ease;
 }
 
+.menu-top {
```
meligatt: @gepifanio I would move this menu top css to the new child component <style> section... `MegaMenuTopNavigation.vue`

gepifanio: Should I add the CSS scoped or in a separate file?

components/page/megamenu/MegaMenu.vue
```diff
@@ -36,10 +37,16 @@
...
-            @click="dismissMobileMenu">Close</div>
+            @click="dismissMobileMenu"
+          >Close</div>
           <PageSearchForm
             v-if="isMobileOpen"
             aria-hidden="true"/>
+          <MegaMenuTopNavigation
+            v-if="topNavigationItems"
+            :top-navigation-items="topNavigationItems"
+            class="menu-top"
```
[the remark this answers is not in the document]
gepifanio: The idea of adding the class in the parent component is because this specific class is only when using the `MegaMenuTopNavigation` in the `MegaMenu` component, so is a scoped class, If I add to the component itself, will mess up when I try to reuse it in another place.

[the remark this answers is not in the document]
meligatt: @gepifanio as discussed in our conversation today, please:
- change the class name menu-top to follow BEM naming convention.
- remove absolute positioning and z-index, and replace the layout with flexbox properties (thee the CSS examples branch )

components/stats/StatsItem.vue
```diff
@@ -1,8 +1,8 @@
 <template>
   <li class="stats-group__item">
-    <div
+    <!-- <div
```
gepifanio: Just to clarify, what is this stats section?

meligatt:
I will close this PR because no time in house for integration with the legacy mega menu CSS. It will outsourced.

PR #942, opened 2019-02-28
Component 1 - Add Faculty Name next to header logo

codeuri:
Modify existing component: (components/page/megamenu/MegaMenu.vue) in  pattern library to display the "faculty name" next to the logo as shown in the mockup above.  If no faculty name is passed to the component, then do not display the element.

Closes FFAM: https://github.com/unimelb/ffam-prototype/issues/1


gepifanio:
Comment: Deploy Started

Building with commit 15259069cc10de87b57ed3931027ea4b4663c75b

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5c7801d8c5e38a00088c2a18

codeuri:
Sorry, initially I was confused by the "Update docs explaining that components must be explicit closing tags to work on Matrix CMS" notification. And then realized that it is related to components only. 
Cheers

components/page/megamenu/MegaMenu.vue
```diff
@@ -16,6 +16,17 @@
...
       </a>
+      <div
+        v-if="title"
+        class="page-title"
+      >
+        <div class="page-title__separator"></div>
+        <a
+          :aria-label="title"
+          class="page-title__text"
+          href="/"
```
meligatt: Hi! @codeuri here some comments   :) 
- According to the technical spec "Faculty name should return you to the Faculty home page", so this href should be a prop `:href` that receives a link.
- this link should be  target `_parent` (open on same browser tab basically)

components/page/megamenu/MegaMenu.vue
```diff
@@ -16,6 +16,17 @@
           >
         </div>
       </a>
+      <div
```
meligatt: @codeuri Sorry but we forgot to add to the spec that this new piece of UI would be better if you can create a new component called `components/page/megamenu/MegaMenuTitle.vue` so please move this markup and CSS to a new component. 
We are working on improving legacy components, like megamenu,  to a more modular approach. I added this details to the zenhub card as well. We are trying our best to not miss any details for you fo next components, thanks!

components/page/megamenu/MegaMenu.vue
```diff
@@ -20,12 +20,12 @@
         v-if="title"
         class="page-title"
       >
-        <div class="page-title__separator"></div>
+        <div class="page-title__separator"/>
```
meligatt: @codeuri yes, you are correct, Vue component need to be closed like `<my-component> ... </my-component>` instead of the self closing one. This is to ensure better integration with the environment that used our pattern-lib "lib" bundle, which is the CMS.

codeuri: Hi @meligatt 
Thanks! I've created a separate component, but I see  Eslinter error when adding component without content `<my-component></my-component>`. 
Could you advise, please?

meligatt: @codeuri You imported correctly, I didn't specified in my description  that using the "kebab-case"  it's for usage in the context of the CMS (matrix), so when they use our vue components, they should use the format `<my-component></my-component>` (which is available because we renamed the component to kebab-case in this file for `webpack` bundle `https://github.com/unimelb/pattern-lib/blob/dev/targets/lib/index.js`) but in the  pattern-lib vue code base we can import a child component like you did: `<MegaMenuTitle title="Faculty name" href="/faculty-name"/>` . sorry for the confusion.

components/page/megamenu/stories/HybridMenu.vue
```diff
@@ -1,7 +1,9 @@
 <template>
   <MegaMenu
     :items="items"
-    active="/random-link" />
+    active="/random-link"
+    faculty-name="This is a <br /> Faculty name"
```
gepifanio: If the `v-html` is to allow the `<br />` tag. When writing the stories, should think that a non-technical user has to understand how to use it, and when he needs to jump to another line, should add `<br />` tag.

gepifanio:
The lint is failing, make sure to run the command 😄 

`yarn lint`
`yarn lint:js --fix`
`yarn lint:css --fix`

components/page/megamenu/MegaMenuTitle.vue
```diff
@@ -0,0 +1,29 @@
...
+  <div
+    v-if="title"
+    class="page-title"
+  >
+    <div class="page-title__separator"/>
+    <a
+      :aria-label="title"
+      :href="href"
+      class="page-title__text"
+      v-html="title"/>
```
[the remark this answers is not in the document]
meligatt: @gepifanio because the original idea was to add a block of text like this: `This is a <br /> Faculty name` to allow to add break lines in specific parts of the words. Now I realised that it's announced incorrectly with screen readers. so I'll request changes.

@codeuri This directive v-html won't be useful for this purposes because screen readers read the string including the html tags. Instead of v-html, could you update this component to render the title props like discusses above `{{ title }}` and add a `width` property for the CMS team to pass a value so they can manage where the faculty name breaks.

components/page/megamenu/MegaMenu.vue
```diff
@@ -16,6 +16,9 @@
           >
         </div>
       </a>
+      <MegaMenuTitle
```
gepifanio: Should the `v-if` verification be used here to not render the component when there is no `facultyName` passed instead of inside of the component?

components/page/megamenu/stories/HybridMenu.vue
meligatt: @gepifanio yeah, we have decided to not use it. Introduce problems with screen reader. cheers.

components/page/megamenu/MegaMenuTitle.vue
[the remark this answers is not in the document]
codeuri: Hi @gepifanio and @meligatt 
- v-html removed
- {{ title }} - added
-width property added

Cheers

components/page/megamenu/MegaMenu.vue
codeuri: @gepifanio 
Lint issues fixed.
Should I remove v-if in Title component?

Thanks

meligatt: @codeuri thanks for the width changes!, I moved the v-if to the parent component (megamenu) just to follow what we've been doing in terms of conditional render, but I don't mind that some components hide itself. In this case I think was ok because the megamenutitle won't be used in other contexts than the megamenu but anyway, I have refactored a bit to move it to done ;)

@gepifanio thanks for the suggestion, 👍

components/page/megamenu/MegaMenuTitle.vue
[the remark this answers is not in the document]
meligatt: @codeuri no worries, I'll update it now. all good :) thank you..

PR #945, opened 2019-03-05
Generic Card Component

codeuri:
Hi @meligatt and @gepifanio 

Worked on Generic Card Component component that handle Discipline/Showcase components.
Created 3 slots for subtitles.
Created and styled slot for links in the footer.
Created stories.
Created tests.
Added styles for  SvgIcon component in title slots.
Added doc.

As you mentioned let discuss 'FFAM theme' here. As I understood we need to add postcss variable and set this variable to card background. Is this correct?

Thanks


gepifanio:
Comment: Deploy Started

Building with commit 62355e7438af97d168d714e82d8e03b6b12aa985

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5c7eab87490bc300075faecf

meligatt:
@codeuri Hello, I think the "override" tasks will have to wait until we actually merge the components from pattern-lib into master so they are available to import via npm dependencies (`@unimelb/pattern-lib-vue"`). I will organize this tomorrow (Thursday) with @TimMansell and add more description to How use the css override system because I see it's not 100% clear how it works.

meligatt:
@TimMansell can we discuss the overrides tasks tomorrow so it's clearer for us how to use it. I think in the case of creating overrides styles from generic card to create: Discipline and Showcase styles will be tricky:
1. Do we import the generic card from pattern lib to FFAM-prototype twice in different component? so it creates Discipline and Showcase? sorry I got confused. let's discuss tomorrow.

TimMansell:
@meligatt, I've documented how to do this in https://github.com/unimelb/ffam-prototype/pull/120.

Create two stories
- One imports default card (discipline story)
- One imports default card and adds a custom class to change background colour (showcase story)

components/cards/GenericCard.vue
```diff
@@ -0,0 +1,62 @@
+<template>
+  <a
```
gepifanio: @meligatt @TimMansell @jaredh-um The whole card should be clickable?

components/cards/GenericCard.vue
```diff
@@ -0,0 +1,62 @@
...
+    <div
+      v-if="cols !== 1"
+      :style="{ backgroundImage: `url(${thumb})` }"
+      class="card__thumb card__thumb--zoom"/>
+    <div class="card__inner ">
+      <h3 class="card__header">{{ title }}</h3>
+      <div
+        v-if="hasSubTitleSlots"
+        class="card__sub-titles">
+        <slot name="sub-title-1"/>
```
gepifanio: @TimMansell is better to keep using `<slot />` or using `props` and we "block" to only certain content being populated into the component?

components/cards/GenericCard.vue
jaredh-um: @gepifanio I would prefer that it is just the image, main title/name, and bottom link(s) that are clickable links (ie; not the card itself and not the description text inside).

codeuri: Hi @gepifanio @jaredh-um 
I've changed the clickable area as mentioned in the above comment.

@TimMansell I've added a custom class prop.

codeuri:
Hi @gepifanio @meligatt @TimMansell 
I've created Stories on FFAM repo but I can't check it as this PR not merged.
Will continue with this after PR Merge.

Issue #894, opened 2019-01-10
[Accessibility] Tabpanel aria-labelledby references non-existent ID


gepifanio:
@anormand commented on [Wed Jan 09 2019](https://github.com/unimelb/find-a-course/issues/1198)

Problem:
- The aria-labelled by attribute for the 'Degrees' tabpanel seems to have an ID that is built dynamically
- After the page loads fully the id is correct (aria-labelledby="ui-tab-34-1") but the initial load is incorrect (aria-labelledby="-1")

Steps to reproduce:
1. Go to the W3C validator: https://validator.w3.org/#validate_by_uri+with_options
1. Validate the following URI: https://study.unimelb.edu.au/find/interests/accounting/
2. Find the following section of code

```HTML
<div role="tablist" class="tabs__tablist max">
	<a tabindex="0" id="ui-tab-34-1" aria-controls="ui-tab-34-panel-1" href="#ui-tab-34-panel-1" role="tab" class="tabs__tab" aria-selected="true">Degrees</a>
	<a tabindex="-1" id="ui-tab-34-2" aria-controls="ui-tab-34-panel-2" href="#ui-tab-34-panel-2" role="tab" class="tabs__tab">Majors/specialisations</a>
</div>
<div tabindex="0" class="tabs__section">
	<div id="ui-tab-34-panel-1" aria-labelledby="-1" role="tabpanel" class="tabs__panel max" style="">
		<div class="course-explainer"><p>The University of Melbourne offers over 270 degrees in a vast range of study areas and specialisations. Browse through the programs we offer in the study area of your choice.</p>
	</div>
```

Solution:
- Change the aria-labelledby attribute to an aria-label 

```HTML
<div role="tablist" class="tabs__tablist max">
	<a tabindex="0" id="ui-tab-34-1" aria-controls="ui-tab-34-panel-1" href="#ui-tab-34-panel-1" role="tab" class="tabs__tab" aria-selected="true">Degrees</a>
	<a tabindex="-1" id="ui-tab-34-2" aria-controls="ui-tab-34-panel-2" href="#ui-tab-34-panel-2" role="tab" class="tabs__tab">Majors/specialisations</a>
</div>
<div tabindex="0" class="tabs__section">
	<div id="ui-tab-34-panel-1" aria-label="Degrees" role="tabpanel" class="tabs__panel max" style="">
[...]

PR #949, opened 2019-03-11
fix(accessibility): tabs aria-label

gepifanio:
#894

### DIFF
--- a/components/tabs/Tab.vue
+++ b/components/tabs/Tab.vue
@@ -5,7 +5,6 @@
     <div
       v-show="isActive"
       :id="`${namespace}-panel-${index + 1}`"
-      :aria-labelledby="`${namespace}-${index + 1}`"
       class="tabs__panel max"
       role="tabpanel"
     >


gepifanio:
Comment: Deploy Started

Building with commit 9930eb10745dacf207ee38bc86110d7b22d53944

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5c85a932a321260008b96730

components/tabs/Tab.vue
```diff
@@ -5,7 +5,6 @@
     <div
       v-show="isActive"
       :id="`${namespace}-panel-${index + 1}`"
```
meligatt: @gepifanio how removing this tag solves the issue?

gepifanio: @meligatt the accessibility checker is complaining that an `aria-labelledby` attribute must point to an ID of one element in the same document, and this one is pointing to one out of the document which doesn't work at all. Also, there is no point to describe an aria-label as the `<slot />` in this case which is a text, do not need to be referenced to an area labeled as the text itself is interpreted for a screen reader.

PR #953, opened 2019-03-13
Focus Wrapper - Hard-code icons

TimMansell:
- Added `<slot>` so a svg `<path>` can be passed down.  If nothing is passed then slot is empty
- Added view-box prop so viewBox can be set
- hard-coded focus box top-left and bottom-right icons using `<path>`
- Imported `<SvgIcon>` directly so CMS can use it rather than relying on `<svg-icon>`

In future, we can investigate refactoring `<SvgIcon>` to import svgs dynamically using `<path>` or base64 in CSS


gepifanio:
Comment: Deploy Started

Building with commit 0d2d961fbfa8c1edd57155fb58755648406e55a0

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5c889ddce96fee000a45ae8b

PR #960, opened 2019-03-14
Comp - 20:  Inset video

codeuri:
- added dynamic sizes
- worked on responsible 
- added stories 
- Unit test for VideoEmbed component
- Unit test for FigureWrap component


gepifanio:
Comment: Deploy Started

Building with commit 794574ffbd65f409bfd56d8ee51eac3fc0a3fc82

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5c8ad12be0c596000c4ccfdc

TimMansell:
@gepifanio, I think it should be in the embed section.

Do you mean changing the storybook embed section name to Video Embed?  There are soundcloud embeds in that section to so a general embed is fine in that case.

TimMansell:
Also, I don't think we require a small option. @jaredh-um can you confirm.

jaredh-um:
@TimMansell Yeah, I wouldn't bother with anything like the small or medium sizes - too small to view. Large only

components/embed/index.css
```diff
@@ -49,4 +49,43 @@
       height: 20rem;
     }
   }
+
+  &--inset {
+    iframe {
```
TimMansell: Don't think we need lines 54-66

PR #961, opened 2019-03-15
Add pre-push git hook script

pmgm:
This script 
- checks that the SemVer is the same as remote
- auto-increments the SemVer pre-release version number
before pushing to the dev branch

### DIFF
--- a/pre-push.sh
+++ b/pre-push.sh
@@ -1,0 +1,62 @@
+#!/usr/bin/env bash
+# pre-push hook for git
+# to activate it
+# ln -s pre-push.sh .git/hooks/pre-push
+# requires
+#   jq, available on homebrew and elsewhere
+#   sponge (from moreutils), available from homebrew and elsewhere
+#   the semver node module
+REPO_ROOT=$(git rev-parse --show-toplevel)
+echo "Git repo is at $REPO_ROOT"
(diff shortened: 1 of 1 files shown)


gepifanio:
Comment: Deploy Started

Building with commit 740bf5cbc81184492e3875774a6a80053c9d5626

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5c8b3f1a45c7de0008caf8ff

TimMansell:
HI @pmgm, does this increment the patch version only and appends -beta?  ie, 3.0.0-beta to 3.0.1-beta

Could you add a brief example describing `auto-increments the SemVer pre-release version number
before pushing to the dev branch` to make sure we all understand it?

pmgm:
Hi @TimMansell here is how it works.

**auto-increments the SemVer pre-release version number before pushing to the dev branch**

If the remote dev branch is the same as the local dev branch and there are any commits to push, this pre-push hook will:
1. increment the _version_ in the package.json file using the command
`semver -i prerelease --preid 'beta'` <current_SemVer>
This means that version 3.0.0 will become 3.0.0-beta.0
A subsequent push will make it 3.0.0-beta.1 and so on.
The hook also creates a commit for the version number update, which will be pushed with the other commits.

Having an updated package number will allow the dev branch to be built with the new version on npm with each branch update, similarly to production.

https://www.npmjs.com/package/semver.

PR #963, opened 2019-03-15
Comp - 7: Pathfinders black option extra class

dyarmachok:
- Create an inverted instance of Contextual Link > No Image Pathfinder with white background
- Add this component to storybook all the cases and possible columns.
- No more than 4 per rows.
-The design must change depends on how many columns you pick as shown in the design specs

Accessibility:
- Make sure is comply with the accessibility standards WCAG 2.1
- the image should accept also an alt text
- the button must be accessible by keyboard

- add unit testing


gepifanio:
Comment: Deploy Started

Building with commit e866f95ab85dd5dbb8fa2af3b3c186f14efa9eaf

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5c8bc5b973a44d0008044cfa

components/cards/CardLink.vue
```diff
@@ -1,18 +1,23 @@
 <template>
   <a
     :href="href"
-    :class="{'bg-inverted': inverted, 'bg-white' : !inverted}"
+    :class="{'bg-inverted card--bdr': inverted, 'bg-white card--bdr-grey' : !inverted}"
```
gepifanio: Hello @dyarmachok,
Consider moving this verification to a computed prop, that way we keep the template clean, and the logic in the `script` section.

Cheers

components/cards/_card-link.css
```diff
@@ -40,10 +40,35 @@
     max-width: 100%;
     margin: auto;
     padding: var(--vr);
-    color: #fff;
     text-align: center;
   }
 
+  .title-inverted {
+    color: #000;
```
gepifanio: Hey @dyarmachok,

Items such as color, try to use the `vars.css` file to store, which makes easier to change later on.

Cheers

components/pathfinder/stories/Link/Story6.vue
```diff
@@ -0,0 +1,39 @@
+<template>
+  <SectionWrap>
+    <div class="grid grid--4col grid--center">
+      <ListItem>
+        <card-link
```
gepifanio: Hey @dyarmachok,

Also in the story is not clear how to use the components as a Vue Component eg: 

If a user wanna use this `card-link` instead of copying the markup but actually passing props through the component:

Card Link component works like that:

to create a list you need to wrap in a component called `<ListItem>` and in your `<card-link>` component you pass the props as you want eg: `<card-link
          :thumb="false"
          :cols="4"
          title="Test 2"/>`

This sort of explanation.. 

Here is a good example to check: [In page Navigation](https://dev--pattern-lib-unimelb.netlify.com/?selectedKind=Page%20%2F%20Navigation&selectedStory=In%20page%20Navigation&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel)

meligatt: @gepifanio i'll create a separate card to update what you mentioned here. I need thi sto get into dev tonight. cheers.

PR #970, opened 2019-03-19
Add checkCommits guard function

pmgm:
A modification of the pre-push script

We want NO local commits when pushing to dev (or $protected_branch)
and ask to confirm.

We will make an exception to the 'don't push to dev branch' rule if:
1. you are on a clean local dev branch
2. you are pushing to the remote dev branch
3. you want **only** to trigger a new npm build with a new version number.

This hook will check that remote and local are the same. That is, same version number and no local commits pending.
Then it will increment the version number, commit and push with an 'updated version number' message

This will trigger the dev build as usual, but the incremented version number will create a new npm version of pattern-lib too.

### DIFF
--- a/pre-push.sh
+++ b/pre-push.sh
@@ -1,6 +1,9 @@
 #!/usr/bin/env bash
 # pre-push hook for git
-# to activate it
+# updates the pre-release version number before pushing
+# This will only have any effect when pushing to the dev (or $protected) branch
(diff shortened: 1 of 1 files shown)


gepifanio:
Comment: Deploy Started

Building with commit 837df1c30d6b3a944f2e888dc42f3c9f9db0e82e

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5c9076b4ec4d240008b56e8f

PR #1024, opened 2019-04-12
refactor(css): variables and font sizes

gepifanio:
Related to the typography and colors override for FFAM.

### DIFF
--- a/components/_vars.css
+++ b/components/_vars.css
@@ -125,7 +125,6 @@
 
   /* FFAM VARIABLES */
   --col-focus: #d8d8d8;
-  --col-background: #fff;
   --col-font: #4a4a4a;
   --col-line: #505d68;
   --col-box-shadow: 0, 0, 0, .12;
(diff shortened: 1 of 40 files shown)


gepifanio:
Comment: Deploy Started

Building with commit b308447589fa885cf98a68d4a1be4a226536d088

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5cb034ed0c0d700008236306

components/_vars.css
```diff
@@ -125,7 +125,6 @@
 
   /* FFAM VARIABLES */
```
TimMansell: Are these FFAM variables or can they be used in other websites?

gepifanio: This section was introduced at the beginning of FFAM project, when we were not sure if create a new repo for FFAM or just do it in `pattern-lib` so to build some components with the `FFAM` styling need that variables, could be a good idea to remove it, and move to `ffam-prototype`.

gepifanio:
> Hi @gepifanio, what is the purpose of these changes? Just updating hard-coded colours to use the variables?

Yes, pretty much this.

PR #1035, opened 2019-04-24
feature(component): headings

gepifanio:
Ticket: [#3](https://github.com/unimelb/ffam-prototype/issues/3)

TODO:

- [x] Use vertical rthym for margins


gepifanio:
Comment: Deploy Started

Building with commit ed6cf03d472d512453ec1e281b1f76f3ae48ef1b

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5cbfd77261f7000008a7fe26

components/typography/headings/Heading1.vue
```diff
@@ -0,0 +1,21 @@
+<template>
+  <div>
+    <h1 :class="size">{{ text }}</h1>
```
TimMansell: Make the component prepend the `title--`. Something like:

```code
 <h1 :class="`title--${size}`">{{ text }}</h1>
```

TimMansell: Also best practice for BEM requires the classes be like:

```code
<h1 :class="`title title--${size}`">{{ text }}</h1>
```

TimMansell: http://getbem.com/naming/

components/typography/_titles.css
```diff
@@ -9,27 +9,30 @@
 .title--xl {
```
TimMansell: Update to use nested `.title { ... }` so it's easier to make changes later on.

TimMansell: `.title` class should have styles same as title--md also

components/typography/_titles.css
```diff
@@ -9,27 +9,30 @@
...
   margin-bottom: calc(2 * var(--vr)) !important;
```
TimMansell: Ideally, we want to not use any `!important` at all throughout the codebase.  It can potentially make it really hard to maintain.

Can you remove all these?  Was there a reason for adding them?

components/typography/headings/Heading2.vue
```diff
@@ -0,0 +1,21 @@
+<template>
+  <div>
```
TimMansell: Since it's a single element being rendered we don't need the wrapping `<div>`

components/typography/_titles.css
gepifanio: I am not sure why I put !important, I tested before and was not working without it, was getting the tag style instead I reckon, when I test again I let you know exactly.

PR #1039, opened 2019-04-29
fix(component): in page nav css

gepifanio:
PR to fix css using BEM and make it possible to override in FFAM

### DIFF
--- a/components/page/navigation/InPageNavigation.vue
+++ b/components/page/navigation/InPageNavigation.vue
@@ -2,14 +2,15 @@
   <FocusWrapper
     :color="color"
     padded>
-    <div>
+    <div class="in-page-navigation">
       <h2 class="in-page-navigation__title">{{ title }}:</h2>
       <hr class="in-page-navigation__line">
-      <div class="in-page-navigation in-page-navigation__container">
+      <div class="in-page-navigation__container">
         <li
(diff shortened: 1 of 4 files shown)


gepifanio:
Comment: Deploy Started

Building with commit 51b0c30969394cd47046b40c3c6a56dadc023f8d

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5cc65483ec440b0008326f7d

components/page/navigation/InPageNavigation.vue
```diff
@@ -3,20 +3,20 @@
     :color="color"
     padded>
     <div>
```
TimMansell: this `<div>` should have then block class `<div class="in-page-navigation">`

components/page/navigation/OutPageNavigation.vue
```diff
@@ -3,20 +3,20 @@
     :color="color"
     padded>
     <div>
```
TimMansell: this `<div>` should have then block class `<div class="in-page-navigation">`

components/page/navigation/index.css
```diff
@@ -0,0 +1,39 @@
+.in-page-navigation {
+  display: flex;
```
TimMansell: Migrate styles to `&__container`

PR #1083, opened 2019-05-15
Card link interactivity

TimMansell:
Generic card:
heading & subtitles underlines on hover
links 1 rem from the bottom.  `card__inner` padding-bottom: 1rem

Staff cards:
Staff name underlines on hover


gepifanio:
Comment: Deploy Started

Building with commit 86e12acba7fa1f9aef06bb97faaa79474b4dbbf7

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5cdb720f966e7d000895f9bf

components/cards/_card-generic.css
```diff
@@ -69,6 +74,10 @@
       font-weight: var(--fw-semibold);
       text-decoration: underline;
```
gepifanio: I think we can remove the underline prop of the `sub-titles` of the cards

components/cards/_card-generic.css
```diff
@@ -69,6 +74,10 @@
...
       color: rgb(var(--col-background-primary-navy));
+
+      &:hover {
```
gepifanio: And will not be necessary to use this :)

components/cards/_card-generic.css
TimMansell: It's strange, I can't see this line when checking out this branch on my local.  Can checkout this branch and see if you can see it?

components/cards/_card-generic.css
TimMansell: I have removed text underline for the `<a>` links.  We'll still need the `:hover` to remove it thought.

gepifanio: Hmm, ok 👍

jaredh-um:
Looks good. Might need to look at those link colours for the UoM version but we can do that down the track.

Issue #1113, opened 2019-05-23
Minify CSS bundle


TimMansell:
## Expected behaviour

ui.css to be minified.

## Current behaviour

ui.css is not minified

## Steps to reproduce

1. https://d2glwx35mhbfwf.cloudfront.net/v4.6.3/ui.css

PR #1152, opened 2019-06-06
chore(perfomance) minify css bundle

TimMansell:
#1113 

Build process outputs an unminified css (to not break any current setups) and now outputs a `ui.min.css` file that is minified.
Updated `yarn build:lib` to handle building of unminified and minified css bundles so semaphore doesn't need to be updated
Installed `npm-run-all` so multiple commands in one script can also be run on windows machines


gepifanio:
Comment: Deploy Started

Building with commit c5a332d641892bad9f0f8b8fded5aabe4620f90e

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5cf8ad736cf4f7000807019a

gepifanio:
Hey Tim, I noticed that is missing some CSS from the unminified one, for eg: `pswp = photoswipe` CSS.

Is that expected to happen?

TimMansell:
@gepifanio, no the photoswipe css should be included.  Thanks for finding that. I will investigate.

PR #1066, opened 2019-05-10
feature(component): carousel 22

codeuri:
- slider
- slider bar
- timing
- tests
- stories
- docs

PR #1183, opened 2019-06-28
chore(component): carousel 22 

codeuri:
fix issues from https://github.com/unimelb/pattern-lib/pull/1066


gepifanio:
Comment: Deploy Started

Building with commit 19dc2b422e02b57829c287a0ab124bb7987b8cad

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5d162fe0bc47ed000780ef6b

components/carousel/Carousel.vue
```diff
@@ -0,0 +1,297 @@
...
+      <div
+        :style="{ width: progressBarWidth + '%'}"
+        class="carousel__panel__progress-bar"/>
+      <div class="carousel__panel__container">
+        <div
+          class="carousel__panel--right"
+          role="navigation"
+          aria-label="Stories"
+        >
+          <div
```
TimMansell: Migrating from other PR:

- Should be ```<ul>``` and have a role of menu, i.e.

```<ul role="menu">```

File: components/carousel/Carousel.vue [hunk 2]
```diff
@@ -0,0 +1,297 @@
...
+            class="carousel__panel--stories-menu"
+            role="menu">
+            <a
```
TimMansell: Should be wrapped in an `<li>`

TimMansell:
Hi @codeuri, some of the tests seem to be broken.  Can you have a look, thanks.

I've fixed some conflicts for you from some other code we merged into `dev` today.

TimMansell:
I've fixed unit tests.

jonathan-doc:
Looking good. I'm still working out typography, so I'll discuss that with Jared first.

For now, just a couple small fixes:
- [ ] The progress bar should have a white background so that it's clear that the function is still present even when the fill clears upon interaction
- [ ] The featured story shouldn't be scrollable; this might confuse readers as to the purpose of the carousel

Also, a single/dual story option isn't necessary, as ideally there would always be three (and it would definitely defeat the purpose of having a carousel to have just the one story).

PR #1189, opened 2019-07-03
feat(component): tags - news tag

gepifanio:
Ticket: [#321](https://github.com/unimelb/ffam-prototype/issues/321)

Be aware,

This PR is changing all the cards styling, as required by the designers.


gepifanio:
Comment: Deploy Started

Building with commit 2af4ec4c91e79a1b57e3d7271ea8e9351dd6d69b

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5d1c33c0d60541000a6b7e6b

components/cards/_card-generic.css
```diff
@@ -18,6 +18,9 @@
   }
 
   .card__title {
+    font-size: 1.25rem;
+    line-height: 1.5625rem;
```
TimMansell: use unitless value

components/cards/_card-generic.css
```diff
@@ -30,17 +33,24 @@
...
+    }
   }
 
   .card__sub-titles {
     border-top: 1px solid rgb(var(--col-black-mid-1));
-    padding-top: 1rem;
+    padding-top: .5rem;
     font-weight: var(--fw-semibold);
+    font-size: .8125rem;
+    line-height: 1.0625rem;
```
TimMansell: use unitless value

components/tags/index.css
```diff
@@ -0,0 +1,28 @@
...
+  white-space: nowrap;
+  flex-wrap: wrap;
+
+  &__item {
+    padding: .125rem .25rem;
+    background-color: rgb(var(--col-blue-light));
+    border-radius: .25rem;
+    font-size: .8125rem;
+    line-height: 1.0625rem;
+    color: #000 !important;
```
TimMansell: use variables where possible

components/tags/Tags.vue
```diff
@@ -0,0 +1,22 @@
+<template>
+  <div class="tags">
+    <a
```
TimMansell: I feel this would be bette suited as al `<ul>`

components/cards/GenericCard.vue
```diff
@@ -22,6 +22,10 @@
         <slot name="sub-title-3"/>
       </div>
       <div class="card__excerpt">{{ excerpt }}</div>
+      <Tags
```
TimMansell: I think we should nest `card__tags` & `card__links` inside a `card_footer` and use flex box to align.  Then we won't need the .`card__tags` css to override some styles.

gepifanio:
> General Comments:
> 
> This new `card news` component looks very similar to `card events horizontal` component. Can you combine them together. Using a `card__footer` should solve most of it.
> 
> There are a few font typography/margin changes. Were those changes requested from design?

Fixed the be using `card__footer`

Yes, the changes were requested by @jonathan-doc, and as we talked to be applied for all the cards.

Just to make sure, @jaredh-um and @jonathan-doc 
Could you guys review the changes, as affected all the cards such as:

`font-sizes`
`paddings`
`icons size`
`colours`

jonathan-doc:
Hey @gepifanio, just some fixes:

**General**
- [ ] The **line-height** for the subtitle and excerpt have been correctly set to 17px (1.0625rem), but they are being computed otherwise; please investigate
- [ ] Set the width of **card__links** to **max-content** so that when tags begin to stack vertically, the link chevron and text are not separated
- [ ] Set the vertical spacing between individual tags to **8px** (they've already been set to 8px horizontally)
- [ ] Set the excerpt's **margin-bottom** to **16px**, up from 8px

**Column / Mobile / Vertical**
- [ ] Reduce the margin between the tags and card link to **12px**

Thanks!

components/cards/_card-generic.css
```diff
@@ -30,17 +33,25 @@
...
```
TimMansell: use unitless value which becomes a computed output of 1.0625rem;

components/cards/_card-events.css
```diff
@@ -1,20 +1,14 @@
...
-    font-weight: var(--fw-bold);
-    text-decoration: none;
-  }
-
   .card__sub-titles {
     margin-bottom: .25rem;
     border-bottom: 0;
     padding: .5rem 0;
 
     svg {
```
TimMansell: The svg height and width will be inherited from the `_card-generic.css` so this becomes redundant.

components/cards/_card-generic.css
gepifanio: In this case, the unitless value was changing the output expected (17px) for some reason. So I added back the rem prop which fixed the output value back to `17px`.

jaredh-um: Unitless line-height values rely on the font-size property to determine the calculation. So for example, a line height of 1.5 on text that is 16px would be 16 * 1.5 = 24px line height.

In the code, if the font size is .8125rem (which equates to 13px when body font-size is 16px) multiplied by a unitless line-height of 1.0625 is
13 * 1.0625 = 13.8125px

PR #1225, opened 2019-08-05
refactor(component): social list bar to be reusable

gepifanio:
Make the social list bar reusable so we can use in other components as well such as Event Header https://zeroheight.com/17lwnkf17/p/1716a8/i/3066168

PR #1222, opened 2019-07-31
feat(component): event info / sidebar

gepifanio:
Component: https://zeroheight.com/17lwnkf17/p/69797d/i/3067630

PR #1226, opened 2019-08-05
feat(component): header text with cta and social list

gepifanio:
To see the button update and social list-icon Waiting to merge first:

#1225 
#1222


gepifanio:
Comment: Deploy Started

Building with commit 8bfb3637be6332b79fb49717fc99c32d6df3d872

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5d47cceae6548e0007cd6b73

components/header/PageHeaderText.vue
```diff
@@ -0,0 +1,69 @@
...
+          width="fullwidth"
+          no-icon>{{ buttonText }}</ButtonIcon>
+        <SocialListGeneric
+          facebook-link="www.unimelb.edu.au"
+          twitter-link="www.unimelb.edu.au"
+          linkedin-link="www.unimelb.edu.au"
+          instagram-link="www.unimelb.edu.au"
+          vimeo-link="www.unimelb.edu.au"
+          youtube-link="www.unimelb.edu.au"
+          class="page-header-text__social-list"/>
```
TimMansell: Don't think we need this as there's no css styling for it?

TimMansell:
`max-width` doesn't appear to be working as intended in IE11. The main title doesn't break onto two lines.

Screen Shot 2019-08-08 at 11 32 51 am

@jaredh-um, @jonathan-doc: should we bother about this?

components/header/stories/text/Default.vue
```diff
@@ -0,0 +1,18 @@
+<template>
+  <PageHeaderText
```
TimMansell: I'd be inclined to call this `<PageHeaderEvent>` as it feels quite specific to an event, rather than generic text.

components/header/PageHeaderText.vue
```diff
@@ -0,0 +1,69 @@
+<template>
+  <div class="page-header-text">
+    <div class="page-header-text__inner">
+      <div class="page-header-text__header">
```
TimMansell: I would remove the wrapping `<div class="page-header-text__header">` and instead use `<h4 class="page-header-text__header">...</h4>`

components/header/PageHeaderText.vue
```diff
@@ -0,0 +1,69 @@
...
+      type: String,
+      default: '',
+      required: true,
+    },
+    descriptionText: {
+      type: String,
+      default: '',
+      required: true,
+    },
+    buttonText: {
```
TimMansell: any props that aren't required should be hidden/shown using `v-if`

components/header/PageHeaderText.vue
gepifanio: Yes true! Thanks, I forgot to remove it, I was styling it before to use it in the generic mode, but now don't need anymore

jaredh-um:
@TimMansell re: the IE11 issue, can you try adding a `width: 100%` property to the `.page-header-event__inner` class (in addition to its existing `max-width` property.

That seems to resolve the issue in IE11 for me (followed [this guide on Stack Overflow](https://stackoverflow.com/questions/35111090/text-in-a-flex-container-doesnt-wrap-in-ie11) [the one that mentions the grandparent]).

PR #1242, opened 2019-08-14
chore(filter): acccordion update data structure

gepifanio:
Update of the accordion filter to accept a new data structure such as:
https://finearts-music.unimelb.edu.au/_web_services/people-listing-via-funnelback/breadth-subjects/dev/id-dev?id=75806


gepifanio:
Comment: Deploy Started

Building with commit 34274a141593ba3b9725ec8bf119dac1fd251b19

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5d53894ebbd0a8000c161b40

components/filter/accordion/AccordionFilter.vue
```diff
@@ -85,7 +85,12 @@
                 <td>
                   <strong>Location</strong>
                   <br>
-                  {{ item.location }}
+                  <span
+                    v-for="(item, index) in item.location"
+                    :key="index">
+                    <span v-if="index > 0">{{ ', ' }}</span>
+                    {{ item }}
```
TimMansell: Might be worth converting line 91-92 into a computed prop that works out the `,`.  Then we can use that as a generic function for use in the generic fitlers (when we get around to it)

components/filter/accordion/accordion.md
```diff
@@ -3,28 +3,36 @@
...
+            "february"
+        ],
+        "location": [
+            "location 1",
+            "location 2"
+        ],
+        "points": "points 1",
+        "discipline": "discipline 1",
+        "audition": "audition 1",
+        "abrod": "abrod 1",
```
TimMansell: I noticed this typo.  Should be `abroad`

components/filter/accordion/AccordionFilter.vue
gepifanio: instead, I added the built in JS array.join method:

`Array.join(', ')`

What do you reckon?

PR #1251, opened 2019-08-21
test vue-agile implementation

gepifanio:


gepifanio:
Closing PR as vue-agile is not compatible with SSR

Issue #1309, opened 2019-10-08
Update Showcase filter VUE.js to include tags


dkevey:
Showcase results under vue.js filter need to be updated to include the tags. Currently using the News Card layout with tags as the event layout hasn't been updated. 
 
https://finearts-music.unimelb.edu.au/showcase
![image](https://user-images.githubusercontent.com/1305804/66358990-f383bb00-e9c0-11e9-9dfb-ce7125d80e21.png)

![image](https://user-images.githubusercontent.com/1305804/66359002-00081380-e9c1-11e9-93d6-adab3a7333e4.png)

PR #1311, opened 2019-10-09
chore(filter): cards update to accept tags

gepifanio:
Ticket: #1309

### DIFF
--- a/components/tags/Tags.vue
+++ b/components/tags/Tags.vue
@@ -2,14 +2,16 @@
   <div class="tags">
     <ul class="tags__list">
       <li
-        v-for="item in data"
-        :key="item.id"
+        v-for="(item, index) in data"
+        :key="index"
         class="tags__item list-reset">
         <a
+          v-if="item.href"
           :href="item.href"
           class="tags__link">
-          {{ item.value || item }}
+          {{ item.value }}
         </a>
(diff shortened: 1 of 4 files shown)


gepifanio:
Comment: Deploy Started

Building with commit 2d63cbd8c0e22f4408873d349d66c4cdaa24dca3

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5d9d5098e2b83d000bb4276d

components/tags/Tags.vue
```diff
@@ -8,7 +8,7 @@
         <a
           :href="item.href"
           class="tags__link">
-          {{ item.value }}
+          {{ item.value || item }}
```
TimMansell: There would be a11y issues if we have an `<a>` without it linking anywhere (href).

I suggest updating `Tags.vue` to something like:

```
<li
  v-for="(item, index) in data"
  :key="index"
  class="tags__item list-reset">
    <a
      v-if="item.href"
      :href="item.href"
      class="tags__link">
        {{ item.value }}
    </a>
    <span v-if="!item.href">{{ item }}</span>
</li>
```

and moving the `padding: .125rem .25rem;` from `&__link` to `&__item`

I updated it to use `index` instead of `item.id`

Issue #902, opened 2019-01-16
Add a11y check in unit test


meligatt:
sources: 
- https://www.deque.com/axe/integrations/
- https://github.com/dequelabs/axe-core/blob/develop/doc/examples/jest_react/link.test.js

PR #1312, opened 2019-10-09
chore(testing): add a11y test to all components

gepifanio:
Ticket: #902 

That's an important PR to make sure all components follow all accessibility requirements.
Could have some breaking changes as listed below:

- Updated some props to be `required: true` as accessibility complain if ever that prop comes empty, that could impact any usage of the component that is not sending the required prop.
- a couple of `<li>` tags were failing with not being wrapped in `<ul>`
- Some of the default `a` tags that were being set before, has been changed to must accepting a valid `href` link

### Tests that has been skipped:
- [x] Accordion
- [x] ToggleBlock


gepifanio:
Comment: Deploy Started

Building with commit 1bea1f18f8efdd0d2516cbd97d381b3f843b5ccc

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5d9d7aa5510e170008ae8116

components/side-panel/SidePanelNavItem.vue
```diff
@@ -1,23 +1,25 @@
...
-        <slot />
-        <SvgIcon
-          class="push-icon__icon"
-          name="chevron-right"
-          width="10"
-          height="10" />
-      </span>
-    </a>
-  </li>
+  <ul class="list-reset">
```
TimMansell: While this now passes the a11y test there are now two `<ul>` when this component is nested inside `SidePanel.vue`

Is there a way to *ignore* the `ul`?

components/stats/StatsItem.vue
```diff
@@ -1,14 +1,16 @@
...
-    <div
-      class="stats-group__icon"
-      v-html="icon" />
-    <strong class="stats-group__value">
-      <span class="stats-group__value__hash">#</span>
-      <span class="stats-group__value__num">{{ number }}</span>
-    </strong>
-    <span class="stats-group__text">{{ meta }}</span>
-  </li>
+  <ul class="list-reset">
```
TimMansell: While this now passes the a11y test there are now two `<ul>` when this component is nested inside `StatsGroup.vue`

Is there a way to *ignore* the `ul`?

TimMansell:
Fixed accordion and toggleblock tests

gepifanio:
Hello, @dkevey 

This had a couple of changes in some of the components, such as:

`Vue Components` now have required props which are compulsory sending otherwise will break the component.

`HTML markup` has changed for example the addition of `aria-label` and `role` changes, so we will need to update the markup. 

Follow attached the list of the components that have changed and might impact you.
[list of components changes.xlsx](https://github.com/unimelb/pattern-lib/files/3732530/list.of.components.changes.xlsx)

components/side-panel/SidePanelNavItem.vue
gepifanio: Humm, thats a good point, I don't think we can detect when this component is nested inside another one to remove the `ul` so probably we will need to change the `a11y` test for this component.

dkevey:
Hi @gepifanio I think it's just the mega menu from this list that is using VUE the rest are using full html. These listed as they are are fine for the release notes - I'll check closer with we update.

image

Issue #1296, opened 2019-09-16
[BUG] News card 1 column layout - Image doesn't show in Safari


TimMansell:
## Expected behaviour

Image should display

## Current behaviour

![Screen Shot 2019-09-16 at 1 46 24 pm](https://user-images.githubusercontent.com/1926461/64933405-80d36400-d888-11e9-8c40-93d18ecfe2b0.png)

## Steps to reproduce

1. https://pattern-lib-unimelb.netlify.com/?selectedKind=Cards%2FNews&selectedStory=1%20Column&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel

### Affected browsers

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] IE 11
- [x] Safari 8+
- [ ] iOS 8.4+
- [ ] Android 4.4+

PR #1332, opened 2019-10-21
fix(component): card news tag image on safari

gepifanio:
Ticket: https://github.com/unimelb/pattern-lib/issues/1296

### DIFF
--- a/components/cards/_card.css
+++ b/components/cards/_card.css
@@ -38,7 +38,6 @@
 
   &__thumb {
     width: 100%;
-    height: 200px;
     background-repeat: no-repeat;
     background-position: center;
     background-size: cover;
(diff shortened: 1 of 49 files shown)


gepifanio:
Comment: Deploy Started

Building with commit b30468a495fe482d035f4bb75a2c94cb3f791c79

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5dad40058b9d0e000989219c

components/cards/_card.css
```diff
@@ -59,10 +58,6 @@
         background: inherit;
       }
     }
-
-    &--full-height {
-      height: 100%;
```
TimMansell: `height: auto;` seems to fix it.

components/cards/_card.css
```diff
@@ -38,7 +38,6 @@
 
   &__thumb {
     width: 100%;
-    height: 200px;
```
TimMansell: need to keep `200px` as this was specified from design team

PR #1360, opened 2019-11-29
chore(component) update styled-select - add ability to pass options object

TimMansell:
`<StyledSelect>` now accepts `options` prop so that can be used instead of default `<slot>` if required.


gepifanio:
Comment: Deploy Started

Building with commit 38c7256d480582aaa1bb94d127a36b9f05a71581

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5de09d76715642000a588cf0

PR #1421, opened 2020-03-16
feat(e2e) snapshot testing

TimMansell:
- Set up e2e visual snapshot testing of components
`yarn test:e2e:snapshots`
`yarn test:e2e:updatesnapshots`

- Initial set up of `buttons` and `carousel`
- All failed tests will have their visual diff located at `cypress/snapshots/__diff_output__`
- Runnning `yarn test:e2e:snapshots` will clean out the `cypress/snapshots/__diff_output__` folder.


gepifanio:
Comment: Deploy Started

Building with commit 7c6bba42c3f3aea7519a3733df892e1213b7d326

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5e6f09eeb63b90000722e1a7

gepifanio:
Awesome PR!

Pretty good, thanks for implementing it.

One question:

As the tests take quite a bit of time to run, would be possible to implement a way to run the test only in the code/components that have been affected by changes.

Eg: I am fixing up an existing component "Accordion" and do some changes to it, and run the snapshots tests, this would run all the tests in the code base, and I just wanna see if the Accordion has changed, because that's the interesting component for me, to see if affected anything. There is a way to implement that? if yes, it might be worth adding the snapshots tests together with the `husky` so it runs and makes sure snapshots have been tests saving time for who reviews don't need to re-run.

TimMansell:
> Awesome PR!
> 
> Pretty good, thanks for implementing it.
> 
> One question:
> 
> As the tests take quite a bit of time to run, would be possible to implement a way to run the test only in the code/components that have been affected by changes.
> 
> Eg: I am fixing up an existing component "Accordion" and do some changes to it, and run the snapshots tests, this would run all the tests in the code base, and I just wanna see if the Accordion has changed, because that's the interesting component for me, to see if affected anything. There is a way to implement that? if yes, it might be worth adding the snapshots tests together with the `husky` so it runs and makes sure snapshots have been tests saving time for who reviews don't need to re-run.

You probably could but it could become quite complex.

- Working out global css that can affect all components would require a full snapshot test
- Working out nested component changes would require a parent component test (ie, Buttons)
- Unintended flow on effects to other components that haven't been touched. For example, a change in the `Accordion` component could actually affect something in the `Carousel` component.  This wouldn't/shouldn't happen (using BEM) but depending on how the CSS has been structured it could.

There could be optimisations made such as:
- only testing some of the component variations (most important)
- each component can have all it's variations tested in one test (one `cy.visit()`) rather than loading a new page for each variation.

PR #1451, opened 2020-05-24
using aria-label instead of aria-labelledby

### LINKED ISSUE #1450: [bug][Accessibility] Aria-labelledby attribute must point to IDs of elements in the same document.
Aria-labelledby attribute must point to IDs of elements in the same document.

This is similar to this issue:https://github.com/unimelb/find-a-course/issues/1842
However the element with the ID does exist.

I suspect that this issue is caused by the incorrect use of **aria-labelledby**.

## Possible solution
The element in question is the icon for the hamburger menu, and I think it can be solved by just using **aria-label** instead, similar to how the search icon is marked up.

mvaldman:
# Description

Aria-labelledby attribute must point to IDs of elements in the same document.

This is similar to this issue:unimelb/find-a-course#1842
However the element with the ID does exist.

I suspect that this issue is caused by the incorrect use of aria-labelledby.

Fixes #1450

## Type of change

# Checklist:

# Reviewer Checklist

- [ ] A11y tested

### DIFF
--- a/components/megamenu/MegaMenu.vue
+++ b/components/megamenu/MegaMenu.vue
@@ -159,7 +159,7 @@
             class="link-icon__icon svg"
             role="presentation"
             focusable="false"
-            aria-labelledby="icon-menu"
+            aria-label="icon menu"
             viewBox="10 10 26 28">
             <path d="M6 36h36v-4H6v4zm0-10h36v-4H6v4zm0-14v4h36v-4H6z" />
           </svg>


gepifanio:
Comment: Deploy Started

Building with commit d010fa75be1941ea01819ba2d7c10599df492afc

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5ecafc73990f6d0006a364d8

PR #1466, opened 2020-07-30
feat(component): button card

gepifanio:
# Description

Fixes https://jira.unimelb.edu.au/browse/VOD-111

preview: https://deploy-preview-1466--pattern-lib-unimelb.netlify.app/?selectedKind=Buttons%2FTypes&selectedStory=Button%20Card&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel

## Type of change

# Checklist:

- [ ] Snapshots tested
- [x] A11y tested

# Reviewer Checklist

- [ ] A11y tested

### DIFF
--- a/components/buttons/ButtonCard.vue
+++ b/components/buttons/ButtonCard.vue
@@ -1,0 +1,56 @@
+<template>
+  <a
+    class="btn-card"
+    :href="link">
+    <div
+      class="btn-card__inner bg-white"
+      :class="colorClass">
+      <div
+        v-if="topLabel"
+        class="btn-card__top-label">{{ topLabel }}</div>
+      <div class="btn-card__label">{{ label }}</div>
+      <SvgIcon
+        class="btn-card__icon"
+        name="arrow-right"
+        width="26px"
+        height="22px" />
+    </div>
+  </a>
+</template>
+
+<script>
+import SvgIcon from '../icons/SvgIcon.vue';
+
+export default {
+  components: {
+    SvgIcon,
+  },
+  props: {
+    topLabel: {
+      type: String,
+      default: '',
+    },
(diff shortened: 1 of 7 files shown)


gepifanio:
Comment: Deploy Started

Building with commit baa4932708a14632295d2eccfa2716cfa7629c2c

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5f22391db108e000081824d8

jaredh-um:
Hey Gui,

These look to be matching the designs really well. A few things:

jaredh-um:
Thanks for updating. Are you able to add some properties like this for the hover and active state animations:

```
.btn-card__icon {
    transition: scale .1s ease-out;
    scale: 1.0;
}

.btn-card:hover .btn-card__icon {
    scale: 1.1;
}

.btn-card:active .btn-card__icon {
    scale: 0.9;
}
```

I'm guessing from the class names that this might impact some other components though?

jaredh-um:
Also I just checked with Andy and the arrow icon can be hidden from screenreaders because they will already know it's a link. He suggests using `aria-role="hidden"`

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute

jaredh-um:
Sorry Gui, just went to preview and there's one more change to get the animation working... I previously had the `scale` property as the one to animate, but since you're using `transform` instead (which is better for browser compatibility) you need to change the animated property to `transform` instead of `scale`.

- [ ] Adjust  `transition: scale .1s ease-out;` to be ` transition: transform .1s ease-out;`

PR #1473, opened 2020-08-04
fix(component): Screen Reader Jump To main content

denisonishchenkou:
These changes are related to [https://github.com/unimelb/homepage/issues/1](url)

- changes the element to button, to make the component work in Firefox
- adds the link role to change the screen reader behavior to work like with default anchor tag
- changes the position of the button to be left aligned

Also I've tested this functionality in Safari and tab navigation works only with an option button pressed (it is a default behavior for safari). When using voiceover everything works as expected.

Could you please review?

### DIFF
--- a/components/buttons/ButtonIcon.vue
+++ b/components/buttons/ButtonIcon.vue
@@ -2,6 +2,8 @@
   <component
     :is="element"
     :class="classes"
+    :role="role"
+    :tab-index="tabIndex"
     :href="isLink && href"
     :aria-disabled="isLink && disabled ? 'true' : false"
     :disabled="!isLink && disabled">
@@ -75,6 +77,14 @@ export default {
     top: {
       type: Boolean,
       default: false,
+    },
+    role: {
+      type: String,
+      default: undefined,
+    },
+    tabIndex: {
+      type: Number,
+      default: undefined,
     },
   },
   computed: {
--- a/components/screen-reader-jump-to/ScreenReaderJumpTo.vue
+++ b/components/screen-reader-jump-to/ScreenReaderJumpTo.vue
@@ -1,11 +1,13 @@
 <template>
   <div class="screen-reader-jump-to">
     <ButtonIcon
-      :href="toElement"
+      element="button"
+      role="link"
+      :tabindex="0"
       no-icon
       size="xsml"
       class="screen-reader-jump-to__skippy screenreaders-only screenreaders-only-focusable btn--inverted"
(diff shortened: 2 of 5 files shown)


gepifanio:
Comment: Deploy Started

Building with commit 443e58be77f0f3d4c5a69afde5325ee3bcddbacd

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5f29551a3f8bfe0007a3d7f3

gepifanio:
- [ ] IE11 first tab focuses it's not working, only if we click on the main screen so it gets focus. As an example, if you try `www.github.com` it works fine.

denisonishchenkou:
Hi @gepifanio  the behavior in FF and IE11 differs from Chrome: In FF focus is triggered on a document and in IE11 focus is triggered on a first element in the body. I couldn't find the universal solution for this case, is it critical? I can try to add event listeners to the document and the first element on a page, but I think this issue can be because of a storybook wrapper. Also I am not sure that it is safe to a add focus event listener to a first element on a page (to fix the issue in IE11). Maybe it's better to keep current behavior? If it's critical I'll work on a document event listener tomorrow. Does this make sense?

Also on github.com focusing works only on a first focus in IE11 and doesn't work in FF.

gepifanio:
> Hi @gepifanio the behavior in FF and IE11 differs from Chrome: In FF focus is triggered on a document and in IE11 focus is triggered on a first element in the body. I couldn't find the universal solution for this case, is it critical? I can try to add event listeners to the document and the first element on a page, but I think this issue can be because of a storybook wrapper. Also I am not sure that it is safe to a add focus event listener to a first element on a page (to fix the issue in IE11). Maybe it's better to keep current behavior? If it's critical I'll work on a document event listener tomorrow. Does this make sense?
> 
> Also on github.com focusing works only on a first focus in IE11 and doesn't work in FF.

Hi @denisonishchenkou, I will check with our Accessibility lead about that and get back to you, I also don't think its a good idea to have an event listener for that.

gepifanio:
Hello @denisonishchenkou ,

I have checked with @anormand our Accessibility lead and it looks good the changes all tests have passed.

The only change that would be good is:

- [ ] Instead of targeting the `id="main"` to target it `role="main"`.
- [ ] Add the documentation explaining that the main `div` need to have a `role="main"` to be able to work.

denisonishchenkou:
I've updated readme and selector. @gepifanio could you please check again?

PR #1483, opened 2020-08-12
Enhance CampaignBanner component in the way to take numerous background sources

eugene-mitnik:
# Description

This PR is created in order to fulfill one of missed requirements
- [x] Banner should accept multiple source images for different screen resolutions/pixel ratio (eg; @1x, @2x, @3x)
@gepifanio @jaredh-um please review

Fixes #3 from "homepage" repo

## Type of change
- [x] This change requires a documentation update (already changed)


gepifanio:
Comment: Deploy Started

Building with commit 5afa975b6836a5877fd17420bbc38ca6ceb56fb0

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5f33d445d960d500075ea5e7

jaredh-um:
Hi @eugene-mitnik 

Just a couple of things...

- in the source code shown in the "Campaign Banner 2021" story, there is a line ` woman diver` - should the second example 1440w image in there be showing as 2x instead of another 1x?

- what are the minimum image requirements for this component to work? i.e; do you have to supply each at least one image for each of the three breakpoints?

eugene-mitnik:
@jaredh-um 
1. Sure it should (see next commit)
2. Only large and medium are mandatory with one source each

package.json
```diff
@@ -26,6 +26,7 @@
     "build:vue": "cross-env NODE_ENV=production webpack --config targets/vue/webpack.config.js",
     "clean:snapshots": "rimraf snapshots/__tests__/__image_snapshots__/__diff_output__",
     "dev": "start-storybook -p 7002 -c .storybook",
+    "dev:public": "start-storybook -p 7002 -h 0.0.0.0 -c .storybook",
```
gepifanio, line 29: Hey @eugene-mitnik 
I am quite unsure of the reason for this addition, could you explain why would be good to have this new option for lunch the development environment in the host 0.0.0.0 ?

components/campaign/CampaignBanner.vue
```diff
@@ -71,5 +78,19 @@ export default {
       high: 768,
```
gepifanio, line 78: Sorry should've said this before, just remembered right now, for the breakpoints variables could you use the file `helpers/viewports.js`

components/campaign/CampaignBanner.vue
```diff
@@ -39,10 +40,16 @@ import ButtonIcon from '../buttons/ButtonIcon.vue';
 export default {
   components: { ButtonIcon },
   props: {
-    background: {
```
gepifanio, line 42: One more thing, you dont need to do only if you feel comfortable doing, but locally the sources of the images which come from './static/` are working fine, but once we deploy it to the server [campaign banner story](https://pattern-lib-unimelb.netlify.app/?selectedKind=Campaign&selectedStory=Campaign%20Banner%202021&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel) the image is broken.

I know we don't have this ability yet to be able to store local images in the repository and we usually only use images that are stored outside of the project folder. But I wanna check if you are able to implement it in the webpack, making the local images available in the server, so the story doesn't get broken once is in there.

gepifanio: The suggestion in the case you would do is adding an `./assests` or `./static` folder

jaredh-um:
Hi @eugene-mitnik some more fixes for this one....

- [ ] add 'max-width: 19rem;' to the `.campaign-banner__heading` class
- [ ] add `display: block;`  to the img element inside the picture (causes a white line underneath otherwise)
- [ ] adjust width value on `.campaign-banner__text` class - set to `15rem`
- [ ] adjust width value on `.campaign-banner .btn--secondary` selector - set to `max-content`

package.json
eugene-mitnik: @gepifanio it was absolutely necessary for tests on smartphone in out local network. It is not a must to leave this command, I just thought it would be handy... So what I should do with it: remove or leave?

components/campaign/CampaignBanner.vue
eugene-mitnik: @gepifanio sure

components/campaign/CampaignBanner.vue
eugene-mitnik: @gepifanio this issue is not about the component or story implementation. I believe it happens during server deploy process and I'm not sure about the way how to resolve it.

package.json
gepifanio: Oh nice, thats good then just leave it ;)

components/campaign/CampaignBanner.vue
gepifanio: Thats ok ;)

jaredh-um:
Looks good 👍

PR #1496, opened 2020-09-01
chore(component): mega menu hover delay

gepifanio:
# Description

Fixes https://github.com/unimelb/design-system-issues/issues/67

## Type of change

# Checklist:

- [ ] Snapshots tested
- [ ] A11y tested
- [ ] Crossbrowser testing (IE11, Safari 8+, iOS 8.4+, Android 4.4+, Firefox ESR (v52.x), iPhone (4s,6), iPad 2, Galaxy s5) 

# Reviewer Checklist

- [ ] A11y tested

### DIFF
--- a/components/megamenu/MegaMenu.vue
+++ b/components/megamenu/MegaMenu.vue
@@ -186,6 +186,7 @@ import MegaMenuTitle from './MegaMenuTitle.vue';
 import MegaMenuTopNavigation from './MegaMenuTopNavigation.vue';
 import Logo from '../logo/Logo.vue';
 import { WIDTH_900 } from '../../helpers/viewports';
+import { TIMER_500 } from '../../constants/timers';
 
 import {
   KEYCODE_TAB,
@@ -245,6 +246,7 @@ export default {
       lastIndex: null,
       isAnimate: true,
       isActive: false,
+      timeOutID: null,
     };
   },
   computed: {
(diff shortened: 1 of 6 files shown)


gepifanio:
Comment: Deploy Started

Building with commit 45ddc1b2316d46351919e7c45f5f3b299d4ac2fc

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5f4d95df1f80bc000838089b

jaredh-um:
Hey Gui,

This is looking great! Well done on adding the timing delay function and the chevrons to indicate a menu item has children, they increase the usability a lot. 👍 

I have some CSS change requests that will help resolve some minor positioning/alignment issues; this is what I've put into the inspector to get the right look, but probably some of the changes might need to be scoped (eg; `.icon` probably is a global).

Screen Shot 2020-09-02 at 1 03 15 pm

- The `.icon` changes are to make the icon slightly bigger, slightly lower and a little further from the text. These values could also be mirrored in the Gen 4 design as they are very similar

- The `menu__link` changes are to make the text links vertically centred, accounting for the 5px border on the bottom

- The `.mega_top` change is to make the top row of secondary links align with the main menu links at the right hand side

- The `font-size` change is just to stop text links bunching up so much between 900px - 1099px breakpoints

I've also spotted an intermittent bug with the menu not disappearing when hovering a link without a menu, but it only happened once and I can't recreate it now. 🤷

jaredh-um (approved):
Great work, behaviour and look is much improved!

PR #1497, opened 2020-09-01
chore(component): remove focuswrapper size options, small styling changes

jaredh-um:
# Description

A function in the generic `FocusWrapper` previously allowed for the selection of small, medium or large focus mark sizes for the wrapper. This resulted in hardcoded width and height values for each of those sizes across all breakpoints, which is not desirable for responsiveness. The new sizing will be controlled by CSS and will result in only one responsive size being available. This sizing can be reviewed if it doesn't work well globally, but the previous sizes don't appear to have been used in many places (and mostly at medium/large).

Other fixes include typography on the two section heading and paragraph component to be in line with other section level components, and other minor typography changes.

## Type of change

This is a potentially non-breaking change if the `focuswrapper` sizing function has been offered in the CMS via a CCT - the component will ignore a size value passed to it and just use the new standard sizing. This means the size option can be removed from CCTs if it is available as it no longer has any effect.

- [X] Bug fix (breaking change which fixes an issue)

# Reviewer Checklist

- [ ] A11y tested


gepifanio:
Comment: Deploy Started

Building with commit bad8494cb7f9dd9e79fd5aed55d68c5712b53ca0

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5f4dd3586b3f780007621bb5

gepifanio (approved):
Awesome job @jaredh-um !!!

PR #1522, opened 2020-10-12
feat(filter): add video icon card filter

gepifanio:
[preview](https://deploy-preview-1522--pattern-lib-unimelb.netlify.app/?selectedKind=Filter%2FComponents&selectedStory=Cards%20Filter%20Categories&full=1&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel)

### DIFF
--- a/components/filter/index.css
+++ b/components/filter/index.css
@@ -78,6 +78,28 @@
     justify-content: space-between;
   }
 
+  .card__thumb {
+    height: 9rem;
+    position: relative;
+
+    @media (--bp-desktop) {
+      height: 8rem;
+    }
+
+    &::after {
+      content: '';
+      background-image: url('~icons/play.svg?fill=#094183');
+      position: absolute;
+      width: 1.25rem;
+      height: 1.25rem;
+      top: 50%;
+      left: 50%;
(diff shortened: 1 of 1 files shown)


gepifanio:
Comment: Deploy Started

Building with commit cdf750446561cf4a2cd67a6c98ce10ff22b1a159

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/5f83b1815a66c900073a6d8e

jaredh-um (commented):
Hey Gui,
This is looking good! Just some small things:

 - can you adjust the background position of the SVG to be `55% 50%` instead of `center` as the shape of triangle makes it look off centre even when it technically is
- can you add a `transition: all ease-out .15s` property to the ::after pseudo element
- can you add a hover element to set `opacity: 1` on the ::after element (the white circle in this case) - I was able to get one when hovering the `card__thumb` itself, but it would probably be better if it happens when you hover anywhere on the card (but the cursor doesn't need to change)

jaredh-um (approved):
Great work! 👍

PR #1556, opened 2021-03-15
Colour tweaks from Today

lmounsey:
# Description

Tweaked colours based on new visual design patterns for homepage work from Today agency.

Summary of changes:
- Changed variable `--col-navy-light` to `237, 242, 245`
- Changed variable `--col-grey-mid-2` to `74, 74, 74`
- Changed variable `--col-grey-mid-3` to `29, 29, 29`
- Changed the default base colour from `--col-black-dark` to `--col-grey-mid-2`
- Remove the default `p` tag styling
- Added default colour of headings + labels of `--col-grey-mid-3`
- Reviewed existing components and updates heading/label style where applicable

The biggest change is the global text colour of `74, 74, 74`. 

Have contrast checked it on the following background colours:

| Variable | RGB value | Contrast ratio |
| --- | --- | --- |
| `--col-white` | 255, 255, 255 | 8.86 |
| `--col-navy-light` | 237, 242, 245 | 7.86 |
| `--col-background-primary-grey` | 247, 247, 247 | 8.27 |
| `--col-background-secondary-grey` | 228, 228, 228 | 6.97 |

The `--col-background-secondary-grey` does not meet the required 7:1 ratio for AAA compliance when 18 points or less, so this might need some tweaking to be slightly lighter.

I've been through components in Storybook and tweaked heading/label colours to use the updated `--col-grey-mid-3` colour. It may still require some review, for e.g. the blockquote could potentially also be updated to `--col-grey-mid-3`  as the new global colour looks a little off when using the `Noto Serif` font.

Fixes # N/A

## Type of change

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist:

- [x] A11y tested

# Reviewer Checklist

- [x] Code is not repeated (DRY)
- [x] ES6+ code is used where possible
- [x] Snapshots tested against a copy of dev snapshots to check UI changes are intended
- [x] A11y tested
- [x] Crossbrowser tested in at least IE11


gepifanio:
Comment: Deploy Started

Building with commit 18202deacae8608df7406b9e2f58339b5fc46759

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/604fee383c6b7b00088cd0b4

gepifanio:
Hey Lucas, 

Looking at the current impact of those changes in the existing components I could find some bugs: 

- https://deploy-preview-1556--pattern-lib-unimelb.netlify.app/?selectedKind=Header%2FLevel%202&selectedStory=Campaign%20banner%20%28no%20logo%29&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel
- https://deploy-preview-1556--pattern-lib-unimelb.netlify.app/?selectedKind=Header%2FLevel%201&selectedStory=Homepage%20campaign%20%28overlay%2C%20.5%20opacity%29&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel
- https://deploy-preview-1556--pattern-lib-unimelb.netlify.app/?selectedKind=Profiles%20%2F%20Accordion&selectedStory=Default&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel

lmounsey:
Thanks @gepifanio - I will fix those up

components/_vars.css
```diff
@@ -20,8 +20,8 @@
   /* #ABBED5 */
   --col-navy-mid-1: 171, 190, 213;
 
-  /* #F2F5F9 */
-  --col-navy-light: 242, 245, 249;
+  /* #EDF2F5 */
+  --col-navy-light: 237, 242, 245;
```
gepifanio, line 24: Hey Lucas, sorry about that but changing it to the `#EDF2F5` doesn't look great in the existing environment, could you use the old colour we had in this case? 

```
 /* #F2F5F9 */
  --col-navy-light: 242, 245, 249;
```

gepifanio (changes requested):

components/_vars.css
lmounsey: I'll talk to Jacob about this one @gepifanio. If we prefer the updated one, would it be possible to add in as a new variable? Perhaps  `--col-navy-light-2`?

lmounsey: I've chatted to design team here and we would like to keep the new colour  - would it be OK if we introduce it as new variable e.g. `--col-navy-light-2` @gepifanio ?

gepifanio: yes sounds good to introduce a new one @lmounsey

lmounsey: Great - I've reverted back `--col-navy-light` and added a new var for us

jaredh-um:
Hi Lucas,

I've reviewed the impact these colour changes have had on all the stories in Storybook. Issues, comments and suggestions below (category naming is per Storybook)

## Adjust colour var
- Rename Navy Light 2 to a grey

The value for this colour was closest in value to Navy Light, so it was suggested as the nearest possible match. However if you want to press ahead with this custom colour, it should be in the `grey` shades as it doesn't have sufficient blue to warrant including in the `navy` shades, which all have a tip to the brand navy. (I know you just renamed this, sorry)

## Inheriting wrong colours, possibly due to removal of default p definition
These components have issues with undesired colours bleeding through, which I suspect is linked to the removal of the global `p` `color` definition.

Rather than going straight to overrides, I think it would be worth re-establishing the global `p` colour (even if it is duplicating the global `html` `color` value), as there seems to be enough specificity on that to be causing issues without it.

#### Cards > Staff List > Division Head
- *Text inside these cards has reverted to general link colour (a blue) - this should be manually overridden to grey mid 2


#### Listing > Image Listing > No padding
- *Paragraph text is link colour - needs to be overridden to grey mid 2


#### Content Block > News block (isolated),
#### Content Block > Combination News and Events
- *Paragraph text for events is link colour - needs to be overridden to grey mid 2


#### Header > Level 1 > Homepage campaign (overlay, .25 opacity)
- BUG: All banner text is grey now. Should be white


#### Sublink menu > 1of2 columns
- *Right hand column paragraph text is link colour now



## Places where new default grey does not look great

These are places where the new base colours have a sub-optimal visual outcome compared to the previous colour(s).

#### Block Quotation,
#### Section > Split Section > Quotes,
#### Testimonials
- currently all grey mid 2, quote marks are a custom grey
- quote text, and quotation marks should change to grey mid 3
- caption/citation can stay as mid 2


#### Filters
- Filters are inheriting body colour for dropdown selects, which is now too soft compared to the surrounding UI. These selects should use darker mid 3 instead


#### Media Gallery
Captions are all in default grey - headings should be changed to the darker grey mid 3

lmounsey:
Thanks for your review @jaredh-um 

I've just committed the following:

- Changed `--col-navy-light-2` to `--col-grey-light`
- Added back default `p` color, tested those components you identified 
- Fixed components where default grey colour looks odd, as per your list

^Only one outstanding is the Filters one - could you show me which one this is? Or add a screenshot?

Thanks!

components/block-quotation/index.css
```diff
@@ -98,11 +98,9 @@ blockquote,
       color: rgb(var(--col-black-light));
     }
 
+    p,
     &__content {
-      &::before,
-      &::after {
-        color: rgb(var(--col-black-light));
-      }
+      color: rgb(var(--col-black-light));
```
jaredh-um, line 103: Hey Lucas, a little bit confused about this section... is it a dupe/nested dupe of the `&__content` section above? "Black Light" is a really light grey and it doesn't appear to be showing through in the [story](https://deploy-preview-1556--pattern-lib-unimelb.netlify.app/?selectedKind=Block%20Quotation&selectedStory=Default&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel)

jaredh-um:
Hey Lucas,

Looking good, lots of little issues fixed.

✅ Adding back `p` global seems to have resolved the colour issues for those affected components

Some minor adjustments: 

- Blockquote - there seems to be a dupe value? I’ve noted this inline as a separate comment (otherwise if needed, change black light to grey mid 3 unless I’m misreading what that’s for)
- Split section with quote - needs adjusting to grey mid 3 as well (eg; `.split-section--withquote .split-section__quote`) (citation is fine to default to grey mid 2)

- Filters (category in Storybook) - see screenshot below - grey mid 2 text + down arrow in dropdown selects should change to the darker grey mid 3 (circled)

Screen Shot 2021-03-19 at 10 53 29 am

components/block-quotation/index.css
lmounsey: I think this is actually for the `.bg-inverted` variation, so `--col-black-light` is actually white here @jaredh-um

lmounsey:
Morning @jaredh-um 

The filter select, and the split section quote should be fixed now.

RE: the blockquote dupe value, I believe the this is the `.bg.-inverted` variant - added a comment above to the other thread

Cheers

lmounsey:
Hi @gepifanio and @jaredh-um 

I've also just added new classes for typography styles. These have changed since the homepage work, and will also be tweaked probably later down the track. I'm keen though to get then into the lib ASAP so that I can start using on the `find-a-course` repo.

I'm happy for you guys to be opinionated on class names / where in storybook it goes. I've opted for `heading-{size}` and `label-{size}` as the class name for now - reason I haven't used a `--` modifier is because these will exist on their own, with out the need for a `label/heading` class (hope that makes sense).

These are the styles/sizes I've added in:

| Class name | Context |
| --- | --- |
| `heading-lg` | Heading large |
| `heading-md` | Heading medium |
| `heading-sm` | Heading small |
| `heading-xs` | Heading extra small |
| `heading-xxs` | Heading extra extra small |
| `label-md` | Label medium |
| `label-sm` | Label small |
| `label-xs` | Label extra small' |

components/block-quotation/index.css
jaredh-um: So it is, I neglected to open the collapsed section above it. As you were @lmounsey

jaredh-um (approved):
Adjustments to those last components re: colour look good. I'm not fussy about the extra type styles/naming... @gepifanio what about you? 

All good otherwise.

lmounsey:
Hey @gepifanio and @jaredh-um 

Is this all good to merge in?

gepifanio (approved):
looks good to me!

Issue #1552, opened 2021-03-09
[Bug] Footer accessibility issues


jaredh-um:
### Currently

- Links in the footer are coloured white and are not underlined. Because the heading/body font colour is also white, there is no way of distinguishing links from body/heading text apart from context cues like icons and placement.

- The placeholder text in the search box does not meet >4.5:1 contrast requirements

- The search input field does not have a border or background that contrasts against the container background at >3:1

### Proposed changes

- [x] **all** links in footer should be underlined by default, with underline removed on hover
- [x] adjust border properties to be on all sides with 2px radius
- [x] adjust search input field placeholder text value to be white, 60% opacity
- [x] add `margin-top` to input field of `0.25rem` to push label up
- [x] adjust padding on input so magnifying glass icon has more padding on left, adjust top value so vertically centred
- [x] adjust `padding-bottom` on input field to `.5rem` to match top padding
- [x] input border on focus should change to white for all sides

### What this should look like (example)

![Screen Shot 2021-03-09 at 1.26.52 pm.png](https://images.zenhubusercontent.com/5acae0cd4b5806bc2bce3338/38f6064d-c85a-40aa-88c9-5e73fe15cdab)

PR #1557, opened 2021-03-16
fix(component): footer accessibility requests

gepifanio:
# Description

Fixes https://github.com/unimelb/pattern-lib/issues/1552

## Type of change

# Checklist:

- [x] Snapshots tested
- [x] A11y tested

# Reviewer Checklist

- [ ] A11y tested

### DIFF
--- a/components/footer/index.css
+++ b/components/footer/index.css
@@ -8,6 +8,10 @@
 
   a {
     text-decoration: underline;
+
+    &:hover {
+      text-decoration: none;
+    }
   }
 
   /* top half with logo, search, popular links, etc. */
@@ -300,6 +304,7 @@
 
     &::placeholder {
       color: rgba(var(--col-background-primary-white), .6);
+      opacity: 1;
       font-weight: var(--fw-light);
     }
   }
--- a/components/footer/index.css
+++ b/components/footer/index.css
@@ -6,6 +6,10 @@
   background-color: rgb(var(--col-brand));
   color: rgb(var(--col-background-primary-white));
 
+  a {
+    text-decoration: underline;
+  }
+
   /* top half with logo, search, popular links, etc. */
   &__top {
     padding: 3rem 0 2.25rem;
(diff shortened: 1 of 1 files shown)


gepifanio:
Comment: Deploy Started

Building with commit a75dd64cf3ca29b247e9c8933ac4a6b34ff263f1

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/605004dec3c8da0007d978b1

components/footer/index.css
```diff
@@ -283,23 +283,23 @@
...
 
     &:focus {
-      border-bottom-color: rgb(var(--col-background-primary-white));
+      border-color: rgb(var(--col-background-primary-white));
     }
 
     &::placeholder {
-      color: rgb(var(--col-blue-mid-1));
+      color: rgba(var(--col-background-primary-white), .6);
       font-weight: var(--fw-light);
```
jaredh-um, line 303: Hey Gui, can you add an `opacity: 1` value to the placeholder element? It might just be Firefox but it looks like it is inheriting a default opacity from somewhere. If you set this to 1, and still use the existing `rgba` value it should look right

components/footer/index.css
```diff
@@ -283,23 +283,23 @@
...
+    transition: border-color .3s;
+    border: 2px solid rgb(var(--col-blue-mid-1));
+    border-radius: 2px;
     background-color: transparent;
     color: inherit;
     appearance: none;
 
     &:focus {
-      border-bottom-color: rgb(var(--col-background-primary-white));
+      border-color: rgb(var(--col-background-primary-white));
```
jaredh-um, line 298: Firefox is showing a `focus-active` state with a blue ring instead of white (Safari doesn't do this) - is there some way to show both at the same time? 
Screen Shot 2021-03-16 at 1 29 49 pm

components/footer/index.css
```diff
@@ -6,6 +6,10 @@
   background-color: rgb(var(--col-brand));
   color: rgb(var(--col-background-primary-white));
 
+  a {
+    text-decoration: underline;
+  }
```
jaredh-um, line 11: Link underlines should be removed on hover

jaredh-um (approved):
Looks good 👍

PR #1560, opened 2021-03-22
fix(bugs): various small issues

### LINKED ISSUE #1559: [Bug/Chore] Various small issues in Gen 2
### Currently

There are a number of minor issues I picked up in `pattern-lib` during a validation check of the base colour changes from Today.

#### Document List
- [x] all icons use an older link colour and don’t match existing colour

#### Error box
- [x] Text does not contrast with background enough. Change orange-dark text to pink-dark

#### Figure
- [x] Caption text is a custom light grey that won’t pass on any background other than white
- [x] Remove custom colours, adjust caption to use new grey mid 2

#### Forms
- [x] Placeholder text - needs `opacity: 1` added to `placeholder` as a global default so the text can be styled accurately at the `color` level

#### Logo > No Padding
- [x] Story and component should not exist - the ‘padding’ is part of the logo and not to be removed (used in footer, not sure where else)

#### Section > Crest
- [x] Story uses a background image at 100% opacity as well as a background colour (navy). The source image is also fully opaque, which means it completely blocks the background colour. The intention was for the background colour to come through so the text is legible.

- [x] An alternative semi-transparent placeholder image should be used so the background colour comes through.

Image used in live CCT: 
- [x] https://cms.unimelb.edu.au/__data/assets/image/0020/3612125/nike-logo-soft.png

- [x] BUG: background image is centred, and when container is expanded wide eventually the background colour leaks around the left and right sides, around the image. This is an undesired effect which should be remedied by changing the source image as per above.

gepifanio:
# Description

Logo no padding removal affected:

- Header
- Footer

Fixes #1559 

## Type of change

# Checklist:

- [x] Snapshots tested
- [x] A11y tested

# Reviewer Checklist

- [ ] A11y tested


gepifanio:
Comment: Deploy Started

Building with commit a825a5c47c5112e9755b7c503e11887a145952da

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/60580aaa0d09090008a8ec56

jaredh-um:
Hey Gui,

Looking good!

One adjustment:
- in `Notice > Danger` story, can you remove the `<p>` from the example inner content as it is causing the text colour to be the body default, which is too dark for this background colour (it should be white).

Other things:
- Forms - the `opacity` change on placeholder text has worked well. We should check this again when the Today colour changes are merged in

- Thanks for getting rid of the 'no padding' logo

jaredh-um (approved):
👍

Issue #1564, opened 2021-03-30
[Bug] Gen 2 hamburger menu closes automatically when search input field entered at certain mobile resolutions


jaredh-um:
### Expected behaviour

Users on mobile should be able to open the hamburger menu, tap into the search field, enter a query and then perform a search, with the menu remaining visible until the search results page is loaded.

![UoM Mobile search.jpg](https://images.zenhubusercontent.com/5acae0cd4b5806bc2bce3338/3f274219-a9e8-47af-b0f8-1f0814bd2cb0)

### Current behaviour

When the search input field is tapped into after opening the hamburger menu, a keyboard briefly appears and then disappears, and the menu closes.

![Mar-30-2021 11-27-33.gif](https://images.zenhubusercontent.com/5acae0cd4b5806bc2bce3338/ae1299eb-e58d-46a4-a1a4-a01cee0c5df1)

...and the following errors are thrown in the console:

![Screen Shot 2021-03-30 at 11.04.09 am.png](https://images.zenhubusercontent.com/5acae0cd4b5806bc2bce3338/21408e9f-9dd3-433c-af7f-3bd46de7922b)

#### MegaMenu.vue component

- The function `closeMobileIfDesktop()` appears to be firing (line 405)
- `closeMobileIfDesktop()` appears to be following its condition `if (this.isMobileOpen)` which contains code to dismiss and close the mobile hamburger menu
- we're not on desktop so it's not clear why this function is firing in the first place, so I've checked what might be firing this
- turns out there is an event listener which is listening for a browser resize on line 268 which is automatically closing the mobile menu on window resize, but only if the browser is perceived to be a "desktop" one (can't resize browsers on mobile generally speaking)
- further investigation reveals "desktop" is not defined anywhere but "mobile" is - line 257
- "desktop" therefore means anything which is not "mobile"
- "mobile" is being defined as ANYTHING with a width measurement of less than 900px
- the Samsung model in question has a raw resolution of 2960 x 1440 pixels (Height x Width)
- because mobiles generally have a very high raw resolution so pixels aren't visible, this "mobile" check is not functioning as intended
- it is unclear why the event listener is triggering a resize event in the first place - perhaps because the keyboard is appearing on screen?

#### Proposed fix

- the check for determining what is a "mobile" needs to be more robust than just checking the current width of the mega menu; many mobiles have higher resolution screens than desktops
[...]

PR #1565, opened 2021-03-30
fix(component): mega menu mobile viewport checker

gepifanio:
# Description

Fixes https://github.com/unimelb/pattern-lib/issues/1564

## Type of change

# Checklist:

- [x] Snapshots tested
- [x] A11y tested

# Reviewer Checklist

- [ ] A11y tested

### DIFF
--- a/components/megamenu/MegaMenu.vue
+++ b/components/megamenu/MegaMenu.vue
@@ -407,9 +407,6 @@ export default {
       }
       if (this.isMobileOpen) {
         this.dismissMobileMenu();
-        this.$refs.panels.forEach((panel) => {
-          panel.classList.remove('open');
-        });
       }
     },
     handleKey(e) {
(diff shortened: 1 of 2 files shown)


gepifanio:
Comment: Deploy Started

Building with commit 9b761ffe54aad7b6ef5e3ef4a2294fbd7289f378

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/60627ed887d08f000776d070

components/megamenu/MegaMenu.vue
```diff
@@ -190,6 +190,7 @@ import MegaMenuTopNavigation from './MegaMenuTopNavigation.vue';
 import Logo from '../logo/Logo.vue';
 import { WIDTH_900 } from '../../helpers/viewports';
```
jaredh-um, line 191: can this line be removed now?

gepifanio: Not really, as we still check the screen width to the menu items, be opened in "mobile" or "desktop" view if there is a desktop with a small size, still should be able to open the menu items like a "mobile"

jaredh-um: Sure. I'm not really a big fan of arbitrary numbers like "900" though; what's that been based on? Nothing from a design perspective. We don't need to change anything here now, but I feel this kind of number should relate to a certain breakpoint or media query, not some random fixed number.

jaredh-um (approved):

components/megamenu/MegaMenu.vue
gepifanio: Yes I agree, we need to improve this sort of things

PR #1580, opened 2021-04-09
chore(css): tidy grey text

### LINKED ISSUE #1558: [Chore] Remove unnecessary custom grey text in components and use global default
### Currently
With [Today's changes](https://github.com/unimelb/pattern-lib/pull/1556) to global colours, the new default type/paragraph colour is a mid grey.

We currently use a similar custom mid grey in several components, which is now no longer required as the new global default is very similar.

#### Cards > Campaign > Default
- paragraph text colour can change to new default so it matches (currently darker)

#### Cards > Generic Cards
- These cards use a custom grey for text. We should replace this grey with the new defaults

#### Section > Two column > Right with round image
- This component uses a custom paragraph colour. Adjust to global default.

#### Profiles > Accordion
- This component uses custom grey for text. We should adjust this to use the global one.

#### Cards > Features Panel
- Paragraph text should be adjusted to new default (grey mid 3) - currently darker mid 2

gepifanio:
# Description

Fixes #1558 

# Checklist:

- [x] Snapshots tested
- [x] A11y tested

# Reviewer Checklist

- [ ] A11y tested

### DIFF
--- a/components/cards/_card-campaign.css
+++ b/components/cards/_card-campaign.css
@@ -3,7 +3,6 @@
     font-weight: var(--fw-normal);
     font-size: 1rem;
     line-height: 1.25;
-    color: rgb(var(--col-grey-mid-3));
     padding: 1rem 1.25rem;
     background: rgb(var(--col-background-primary-white));
     margin: 0 .4375rem;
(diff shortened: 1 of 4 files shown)


gepifanio:
Comment: Deploy Started

Building with commit e8ff5e055019e8d4688922de91d30aa304b826bd

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/606fcd6ff6a0e80007f5a0a9

PR #1582, opened 2021-04-12
fix(colour) heading classes

gepifanio:
# Description

Fixes https://app.zenhub.com/workspaces/global-design-system-issues-tracker-603c265651b123001c916450/issues/unimelb/pattern-lib/1581

## Type of change

# Checklist:

- [ ] Snapshots tested
- [ ] A11y tested
- [ ] Crossbrowser testing (IE11, Safari 8+, iOS 8.4+, Android 4.4+, Firefox ESR (v52.x), iPhone (4s,6), iPad 2, Galaxy s5) 

# Reviewer Checklist

- [ ] A11y tested

### DIFF
--- a/components/forms/stories/InlineSearchForm.vue
+++ b/components/forms/stories/InlineSearchForm.vue
@@ -4,6 +4,9 @@
       Find a Course
     </h2>
     <form class="form form--inline">
+      <label
+        for="search-input"
+        class="heading-card">Search careers, interests, courses or specialisations</label>
       <div class="form--inline__row form--inline__row--center">
         <span class="form--inline__select">
           <label
@@ -17,6 +20,7 @@
           </select>
         </span>
         <input
+          id="search-input"
           aria-label="Course title or keyword"
           class="form--inline__search"
           type="search"
--- a/components/typography/_headings.css
+++ b/components/typography/_headings.css
@@ -41,6 +41,15 @@ h6,
   font-weight: bold;
   text-transform: none;
   letter-spacing: normal;
+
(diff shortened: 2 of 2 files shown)


gepifanio:
Comment: Deploy Started

Building with commit c67d6fc4f0f0888db89d532106c67704de14f76a

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/6073e0234ac46700085c69d1

components/forms/stories/InlineSearchForm.vue
```diff
@@ -17,6 +20,7 @@
           </select>
         </span>
         <input
+          id="search-input"
           aria-label="Course title or keyword"
```
jaredh-um, line 24: Just curious - do we need a `label` and an `aria-label` for the same input (we may do, I just don't know)? I tried to check with SiteImprove and it doesn't seem to be causing issues but it's hard to tell because it's detecting issues from other stories...

jaredh-um (approved):
This all looks good, as long as having both the label types is acceptable... 👍

components/forms/stories/InlineSearchForm.vue
gepifanio: We don't need, but is not wrong to have both, the main reason Ive added was to test the same case as we have in study which was using the `heading-card` class which was black before, and to be able to test in the storybook I have added this new label using the same class :)

jaredh-um: That sounds fine then 👍

PR #1613, opened 2021-05-10
feat(component): quick links alt

lmounsey:
# Description

New `QuickLinksAlt` component from Today. Can be used in a link or button context. Have documented both, as well as in a search results list with a query.

## Component preview
[Quick Links Menu - Links](https://deploy-preview-1613--pattern-lib-unimelb.netlify.app/iframe.html?selectedKind=Today%2FQuick%20Links%20Menu&selectedStory=Links)
[Quick Links Menu - Buttons](https://deploy-preview-1613--pattern-lib-unimelb.netlify.app/iframe.html?selectedKind=Today%2FQuick%20Links%20Menu&selectedStory=Buttons)
[Quick Links Menu - Variations](https://deploy-preview-1613--pattern-lib-unimelb.netlify.app/iframe.html?selectedKind=Today%2FQuick%20Links%20Menu&selectedStory=Variations)

## Type of change

- [x] This change requires a documentation update

# Checklist:

- [x] Snapshots tested
- [x] A11y tested

# Reviewer Checklist

- [x] A11y tested


gepifanio:
Comment: Deploy Started

Building with commit 4b55328f4648d3d1d59ee51dd5e64e7029784352

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/6098c4ac18604e0008926f22

lmounsey:
Hi @jaredh-um 

Use case is where we display this list in an A-Z (used on the Study index) or as search results, so would be preferable to change the markup to an ordered list.

It won't have any effect in term of styling; there is a list reset on there so shouldn't have any numbers etc.

gepifanio (changes requested):
Looks good,

Just please add in the docs an example of story using all props, also in the readme add the code example on how to do that.

Eg: it has a prop called `items` but what is the structure of this array?

Look this example of story showing what items are in this case: https://pattern-lib-unimelb.netlify.app/?selectedKind=Media%20Gallery%2FIn%20Page&selectedStory=Default&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel

gepifanio (approved):

anormand:
### Quick Links Menu - Links

**1. Navigation landmark**

Problem:
- We think of this component as a set of pathfinder links, but this isn't indicated to assistive technologies

Solution:
- Add a role of navigation and an aria-label of 'Quick links'

Change
`<div class="content-block content-block--sml">`

To
`<div role="navigation" aria-label="Quick links" class="content-block content-block--sml">`

**2. Inaccessible SVGs**

Problem:
- The SVGs are decorative but they are given a role of img. As a result the aria-label 'arrow-right' is read out to assistive technologies

Remedy:
- Give the svg a role of presentation and give it an aria-hidden attribute set to true

Example:
Change
`<svg width="0" height="0" aria-label="arrow-right" focusable="false" role="img">`

To
`<svg width="0" height="0" aria-label="arrow-right" focusable="false" aria-hidden="true" role="presentation">`

### Quick Links Menu - Buttons

**1. Navigation landmark**
- See previous comments

**2. Inaccessible SVGs**
- See previous comments

**3. Invalid links**

Problem:
- Links don't have a href attribute and so can't receive focus

Remedy:
- Add a href attribute

Example:
Change
`<a role="button" data-preview="Quick view" class="quick-links-alt__link quick-links-alt__link--btn">`

To
`<a href="#" role="button" data-preview="Quick view" class="quick-links-alt__link quick-links-alt__link--btn">
`
### Quick Links Menu - Variations

- See previous comments

lmounsey:
I've fixed the  **Inaccessible SVGs** and **Invalid links** issues

RE: Navigation landmark

I think this is dependant on the context in which it is used - e.g. when used as A-Z on the study index it probably doesn't need landmarks for every component as it's grouped A-Z.

I've updated the `Links` and `Buttons` docs to include this markup, but left it off the `Variations`.

The navigation role/aria-label also needs to be on a surrounding tag, as the `QuickLinksAlt` component is a `ul` or `ol`

anormand:
### 1. Quick Links Menu - marked content

Problem:
- The [default appearance of marked content](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/mark), yellow background, is overridden and is presented only as bold text
- Users who benefit most from highlighting, such as those with cognitive impairments, are used to using background highlighting in other programs

Solution:
- Add a light-blue background

Example
```
.quick-links-alt__link-title mark {
    font-weight: 700;
    color: #000000; 
    background-color: #CDE6FF;
}
```

**2. Quick Links Menu - marked content background**

Problem:
- SortSite does not like the CSS that is being used, presumably because the background-color is not specified

Solution:
- Add a background-color element to the CSS

Example

Change:
```
.quick-links-alt__link-title mark {
    font-weight: 700;
    background: 0 0;
    color: #1d1d1d;
}
```

To:
```
.quick-links-alt__link-title mark {
    font-weight: 700;
    color: #000000; 
    background-color: #CDE6FF;
}
```

### 3. We need to included the expected screen reader announcement in the Storybook documentation (see below).

Screen reader announcement (NVDA)

Links:
Start your application link


Buttons:
Accounting Quick view button

Variations
marked content Acc out of marked content outing link

anormand:
This now passes accessibility testing. No errors in SortSite or Siteimprove

PR #1622, opened 2021-05-18
fix(component): section with icons to use div wrapper

gepifanio:


gepifanio:
Comment: Deploy Started

Building with commit 0f80f385cf3a0eb6700b4d55fccf9ccece19c1c0

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/60a36777a0ff2d00072d7ac6

components/_utils.css
```diff
@@ -543,6 +543,16 @@ a[target='_blank'] {
   max-width: 28rem;
 }
 
+/**
+ * Icon helper for section alt wrapper
+ */
+.icon-width-height {
```
jaredhenderson, line 549: This naming is too generic, and is not helpful as a utility since this size is uncommon. I would just add something to the existing component that caters for this specific case/size, eg:

```
.section-alt__inner svg {
  width: 2.5rem;
  height: 2.5rem;
}
```
but this is quite broad as it will affect every SVG inside this `.section-alt__inner` container. (You might want to check the [3 Up](https://deploy-preview-1622--pattern-lib-unimelb.netlify.app/?selectedKind=Today%2FSection&selectedStory=3%20Up&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel) story at the same time as it also uses SVGs and will have the same issues)

Another alternative might be a class that can be added to the `div` containing the SVG like you have, but with contextual to the component, eg;

`.section-alt__inner-svg-icon` or something

jaredhenderson (changes requested):

jaredhenderson (approved):
👍 this will work, but we need to be careful that it is quite broad and will affect any SVG in this group. At the moment that it is behaving as we want it to as there is only one size shown.

PR #1634, opened 2021-05-25
fix(component): mega menu chevron icon

gepifanio:
# Description

image

Fixes: Index - 2.4.4 Link Purpose

## Type of change

# Checklist:

- [x] Snapshots tested
- [x] A11y tested

# Reviewer Checklist

- [ ] A11y tested


gepifanio:
Comment: Deploy Started

Building with commit 61d71371f6b63e3be0222e796e20fc4be4875f2c

https://app.netlify.com/sites/pattern-lib-unimelb/deploys/60ac865fc01f46000745eace

components/megamenu/MegaMenu.vue
```diff
@@ -103,6 +105,7 @@
                         <SvgIcon
                           v-if="!isMobileOpen"
                           :name="'chevron-right'"
+                          aria-hidden="true"
```
jaredhenderson, line 108: - [ ] There's another instance on line 141 that probably also needs to be changed

jaredhenderson (approved):
Looks to resolve the issue, though I think we need to take a serious look at how SVGs are implemented in the future as they cause constant issues - if they're being hidden from screenreaders, there appears to be a bunch of other properties that are perhaps no longer relevant (`title`, `aria-label`, `role` etc.)

PR #1648, opened 2021-06-03
fix(accessibility): campaign-banner-alt background contrast

jaredhenderson:
## Accessibility issue
SiteImprove is reporting the header background as white, causing the overlaid text to fail contrast checks as it is also white. The background has been set to dark navy, but this is failing to be set as there is a typo in the CSS var name used.

This is a false positive however, as there is a dark background image plus gradient in the layers between the text and the background, so is highly legible.

Screen Shot 2021-06-03 at 2 00 05 pm

## Solution
- Background colour has been set, but there is a typo in the colour variable name used. This has been corrected and should enable the background colour to be set to dark navy as originally intended, which should resolve this issue.


gepifanio (approved):
good catch!! 🎉

PR #1649, opened 2021-06-03
fix(accessibility): Override Normalize styles to browser defaults for Firefox focus rings

jaredhenderson:
## Currently
Screen Shot 2021-06-03 at 2 48 51 pm

The `Normalize.css` included in the pattern-lib build resets the default focus styling for Firefox to a single black dotted line on certain button/input elements. This renders it very difficult to see on dark backgrounds.

Reported as an accessibility issue:

> Problem
> Tabs that have focus do not have borders on the right or the bottom
> SortSite is complaining about the tab borders not being visible
> Solution
> See if the outline-width and button border styles can be made visible

## Solution
I've `revert`ed this Normalize change in our `base.css`, so these properties should come from the Firefox user agent stylesheet (browser defaults) instead.

Screen Shot 2021-06-03 at 2 55 57 pm

### DIFF
--- a/components/_base.css
+++ b/components/_base.css
@@ -187,6 +187,14 @@ textarea {
   line-height: inherit;
 }
 
+/* Remove Normalize override styles for Firefox focus rings */
+[type='button']:-moz-focusring,
+[type='reset']:-moz-focusring,
+[type='submit']:-moz-focusring,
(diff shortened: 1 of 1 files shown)


gepifanio (approved):
looks good, well done!

PR #1656, opened 2021-06-07
fix(component): skip to content and icons injection

gepifanio:
# Description

Fixes https://app.zenhub.com/workspaces/global-design-system-issues-tracker-603c265651b123001c916450/issues/unimelb/pattern-lib/1653

## Type of change

# Checklist:

- [x] Snapshots tested
- [x] A11y tested

# Reviewer Checklist

- [ ] A11y tested


jaredhenderson (changes requested):

components/screen-reader-jump-to/stories/Default.vue
```diff
@@ -11,37 +11,37 @@
...
-        <p
-          v-for="key in 20"
-          :key="key">
-          Lorem, ipsum dolor sit amet consectetur adipisicing elit. Sint explicabo quibusdam commodi quidem vitae
-          quaerat cumque, ex velit facilis. Minus ipsum quidem temporibus doloremque molestias doloribus delectus, nemo
-          necessitatibus ut!
-        </p>
-      </section-wrap>
-    </div>
+    <main id="main">
```
jaredhenderson, line 14: The skip to content is jumping to the right section now. Can you add/adjust the following:
- [ ] Add smooth scroll
- [ ] Adjust story contents so the main content + ID is further down the page so we can test if it is scrolling to the correct spot - it's a bit too close to the top to be sure at the moment (be sure not to break SiteImprove though)
- [ ] Move button to first child after `id="main"` heading so we can check if focus is moved correctly

jaredhenderson (changes requested):

PR #1657, opened 2021-06-07
fix(component): section alt icons

gepifanio:
# Description

Icons in the [Story](https://pattern-lib-unimelb.netlify.app/?selectedKind=Today%2FSection&selectedStory=Standard%20%28with%20columns%29&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel) are broken

Code affects stories:

[3 up](https://deploy-preview-1657--pattern-lib-unimelb.netlify.app/?selectedKind=Today%2FSection&selectedStory=3%20Up&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel)
[Sectio with columns and icons](https://deploy-preview-1657--pattern-lib-unimelb.netlify.app/?selectedKind=Today%2FSection&selectedStory=Standard%20%28with%20columns%29&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel)

## Type of change

# Checklist:

- [ ] Snapshots tested
- [ ] A11y tested

# Reviewer Checklist

- [ ] A11y tested


jaredhenderson (changes requested):
- [x] It's really hard to compare what is going wrong between these two components as they are using different SVG sets. Can you please change the 'Standard - with columns' story to use the same style of SVG icons from Matrix - you should be able to use the Matrix image URL directly as well (?) so the SVG data doesn't have to be hardcoded into the story
- [x] I will review the issue again after this

components/today/stories/Section/StandardWithColumns.vue
```diff
@@ -19,9 +19,10 @@
     <div class="grid grid--lg shim-mt2--desk">
       <div class="cell cell--lg cell--tab-1of2 content-max-width shim-mt0 shim-mb0 shim-mb1--desk">
         <div class="section-alt__inner-svg-icon shim-mb067">
```
jaredhenderson, line 21: - [ ] Remove `shim-mb067` from this div - margin is being set using the `section-alt__inner-svg-icon` class already, don't need this twice

jaredhenderson (approved):
much tidier! 👍

PR #1672, opened 2021-06-23
Fix padding for sections in succession

lmounsey:
# Description

Update the selector to fix padding on sections when two sections with just a white background are used in succession. The previous selector was applying to all sections, see https://find-a-course-preview-staging.netlify.app/find/micro-credentials/cybersecurity-technologies-for-defence/ (broken padding on mobile and reduced on desktop)

The selector is a bit ugly - initially I was trying to be a bit clever and not have to have say a default class of `section-alt--standard` but maybe that is easier. That would require a change though I presume to existing components in Matrix.

Fixes # (issue number)

## Type of change

# Checklist:

- [ ] Snapshots tested
- [ ] A11y tested

# Reviewer Checklist

- [ ] A11y tested

### DIFF
--- a/components/section/_section-alt.css
+++ b/components/section/_section-alt.css
@@ -6,7 +6,7 @@
     background-color: rgb(var(--col-grey-light));
   }
 
-  &[class~='section-alt'] + [class~='section-alt'] &__inner {
+  &:not([class~='section-alt--bordered']):not([class~='section-alt--bg-grey-light']):not([class~='section-alt--iimg']) + &:not([class~='section-alt--bordered']):not([class~='section-alt--bg-grey-light']):not([class~='section-alt--img']) &__inner {
     padding-top: 0;
 
     @media (--bp-wide) {
--- a/components/section/_section-alt.css
+++ b/components/section/_section-alt.css
@@ -6,7 +6,7 @@
     background-color: rgb(var(--col-grey-light));
   }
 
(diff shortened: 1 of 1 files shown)


jaredhenderson:
Hmm, I see what you mean @lmounsey. `section-alt` was added to the top level of each of these sections to get those rules to work, but as a result it's no longer specific enough just for the white background version.

Options would be:
- add specific class for white backgrounds as you mention (requires CMS template updates)
- use a rule to select only the applicable sections for this rule to apply (as you've included)

I'm not sure which is best, I guess it depends if we can settle the issue with the rule alone or whether it will complicate things later?

For the second option, it looks like the rule can be slightly simplified with the string selector * (as in, "doesn't contain this partial string anywhere in the class values":

```
.section-alt:not([class*='section-alt--']):not([class*='section-alt__']) + .section-alt:not([class*='section-alt--']):not([class*='section-alt__']) .section-alt__inner {
 padding-top:0
}
```

lmounsey:
Thanks @jaredhenderson 

I think option 2 might be best in this scenario so that we can move the Microcert work out the door - with potentially a longer term look at option 1.

I like the simplification too - will update the selector. We can probably also remove the `:not([class*='section-alt__'])` part as well as it will only be the top level block class

jaredhenderson:
No worries @lmounsey, sounds good

jaredhenderson (approved):
👍

gepifanio (approved):
looks good

PR #1676, opened 2021-06-25
fix(accessibility): Add left border to notices to meet non-text contrast requirements

jaredhenderson:
- Adds a left border with contrast greater than 3:1 to the notice components in order to comply with [WCAG non-text contrast requirements](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html).

Screen Shot 2021-06-25 at 12 21 59 pm


gepifanio (approved):
looks good

PR #1687, opened 2021-07-07
Fix and improvements - homepage components

gepifanio:
# Description

 
## Video iframe incorrectly titled
The video component takes these parameters (label="Life at Melbourne"):
Graphical user interface, text, application, chat or text message

Description automatically generated
But outputs this iframe code (title="Youtube video)":
Text

Description automatically generated with low confidence
The title attribute of the iframe should reflect the label property

## Campaign Banner requires subtitle
an H1 and a link are not sufficient for Campaign Banners.
eg. in the case of Open Day, we'd want something like:
Open Day
August 10, 2021
[ Join us ]
I have resolved this within our CMS https://www.unimelb.edu.au/?banner=3822735, but the pattern should probably be rolled back into the vue component with a property to generate the subheading.

## Card Article with video
A last minute request saw us add a video to a card in place of an image. (First card in last section of the page)
Because the page is maintained largely as static html, this wasn't terribly difficult and worked surprisingly well apart from a couple of small wrinkles.

1) Aspect ratio of the video is slightly different to an image, with the result that the top of the element, as well as the cta link are slightly misaligned when beside a standard image card. Whether it's a matter of making the sizes consistent, letterboxing the video, or some other solution doesn't matter. There’s also a slight problem in that there’s a 1 pixel border on the right of the small embedded video.

2) In this small format, the SVG controls/labels are rather large. Would be nice if they could be a little more proportionally sized. Especially the ‘close’ button, which could maybe be semi transparent when the video is playing or only appear on hover.

Initially, the video was inserted as standard embed code, but I'm hoping to get the OK to switch to using the video component, (though it would be even better if the other issues regarding video embed components - title and autoplay - could be resolved).
At the moment, to see the vue component version of the video, you need to load https://www.unimelb.edu.au?preview=video
Of course, it then follows that the video option would need to be incorporated into the parent card component, and so on. Complicated, I know.

## Type of change

# Checklist:

- [ ] Snapshots tested
- [ ] A11y tested

# Reviewer Checklist

- [ ] A11y tested

### DIFF
--- a/components/cards/CardFlat.vue
+++ b/components/cards/CardFlat.vue
@@ -18,13 +18,20 @@
       </ButtonIcon>
     </div>
     <div
+      v-if="video"
+      class="card-flat__img">
+      <VideoEmbed
+        video-api
+        :src="video.src" />
+    </div>
+    <div
       v-if="img"
       class="card-flat__img">
       <client-only>
         <progressive-img
           :src="img.src"
           :alt="img.alt || ''"
-          :aspect-ratio="0.6"
+          :aspect-ratio="0.56"
           class="card-flat__img-inner" />
       </client-only>
     </div>
@@ -33,10 +40,12 @@
 
 <script>
 import ButtonIcon from 'components/buttons/ButtonIcon.vue';
+import VideoEmbed from 'components/embed/VideoEmbed.vue';
 
 export default {
   components: {
     ButtonIcon,
+    VideoEmbed,
   },
   props: {
     title: {
@@ -56,9 +65,14 @@ export default {
         return null;
       },
     },
+    video: {
+      type: Object,
+      default() {
+        return null;
+      },
+    },
     img: {
       type: Object,
-      required: true,
       default() {
         return null;
       },
--- a/components/cards/CardFlat.vue
+++ b/components/cards/CardFlat.vue
@@ -21,7 +21,7 @@
       v-if="video"
       class="card-flat__img">
       <VideoEmbed
-        video-api
+        :video-api="video.videoApi"
         :src="video.src" />
     </div>
     <div
--- a/components/embed/VideoPlayer.vue
+++ b/components/embed/VideoPlayer.vue
@@ -56,7 +56,7 @@
       <iframe
         v-else-if="videoPlaying"
         ref="embed"
-        title="Youtube video"
+        :title="label"
         width="560"
         height="315"
         :src="videoSrc"
--- a/components/campaign/CampaignBannerAlt.vue
+++ b/components/campaign/CampaignBannerAlt.vue
@@ -79,7 +79,7 @@ export default {
     },
     subHeading: {
       type: String,
-      required: true,
+      required: false,
       default: '',
     },
     tag: {
--- a/components/campaign/CampaignBannerAlt.vue
+++ b/components/campaign/CampaignBannerAlt.vue
@@ -31,6 +31,11 @@
           </a>
           {{ title }}
         </h1>
+        <p
+          v-if="subHeading"
+          class="heading-sm">
+          {{ subHeading }}
+        </p>
         <ul
           v-if="links && links.length"
           class="campaign-banner-alt__links list-reset">
@@ -68,6 +73,11 @@ export default {
   },
   props: {
     title: {
+      type: String,
+      required: true,
+      default: '',
+    },
+    subHeading: {
       type: String,
       required: true,
       default: '',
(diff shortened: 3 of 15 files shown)


components/cards/CardFlat.vue
```diff
@@ -17,14 +17,21 @@
         {{ cta.text }}
       </ButtonIcon>
     </div>
+    <div
+      v-if="video"
+      class="card-flat__img">
+      <VideoEmbed
+        video-api
```
gepifanio, line 24: remove video-api

woowoowoo:
**Campaign Banner requires subtitle**
This is good, thanks!

**Card Article with video**
Alignment/aspect ratio is good, thanks.

Video controls and Video title still need work.

As in the original description, the directly embedded video currently on the home page is only there because the video vue component has problems. To see how it will ideally end up, view the home page with this query string, so you'll see this:
Screen Shot 2021-07-08 at 9 11 33 am
The play button and label are just a bit too large in this small format video. If they can be adjusted proportionally, that would be great, but it can be left for later.

Once the video starts playing, however, the problem is greater. The 'close' button is proportionally quite large, and opaque, so it significantly obstructs the video.
Screen Shot 2021-07-08 at 9 12 45 am
On YouTube videos, it also partially obstructs the 'copy link' button within the video itself. Not sure it's possible to avoid that, Youtube puts clutter and controls all over the place, so there's nowhere that doesn't obstruct something!
Screen Shot 2021-07-08 at 9 16 23 am

The iframe generated by the vue component does not have a title attribute. This is not just a problem in the card implementation - I think all the video components are failing to add a configurable title attribute. This should be handled much the same way as alt text is for images. The user needs to be able to input text to describe the content of the iframe. WCAG guidelines will be satisfied if that is rendered as a title attribute on the iframe.

As you're aware, a new issue has come up with the Related Videos parameter.
Video embed URLs often pass parameters to the video to control aspects of the presentation. Users need to be able to access and edit those parameters.
In the short term, just add rel=0 to the list of parameters passed to Youtube when the jsapi is enabled, but longer term we need a better solution. Things like the starting time (t=72 to start a video 72 seconds in).

components/embed/VideoPlayer.vue
```diff
@@ -56,7 +56,7 @@
       <iframe
         v-else-if="videoPlaying"
         ref="embed"
-        title="Youtube video"
+        :title="label"
```
gepifanio, line 59: @woowoowoo

woowoowoo: So, label is now a property that a user can edit?

And that property will render as `title="label"` on the resulting iframe?

gepifanio: Label is this one: 

image

woowoowoo: OK, that is not what I mean at all.
The vue component writes the iframe html, right?
The generated iframe html currently does not have a title attribute. It needs one.
The user needs to be able to set that title attribute.

woowoowoo: Separately, though.
The query params added for videoApi should include rel=0.
This means that the related videos shown are from the same account as the original video - not just random sh*t.

jaredhenderson: @woowoowoo from what I can see, I think @gepifanio has just taken the original `label` prop (eg; 'Melbourne Connect' in one of your examples above), and is re-using that value to also populate the iframe `title` instead of hardcoding 'Youtube video'. So in the same example, the iframe would be `title`d 'Melbourne Connect'.

As it stands in this PR, this means a user does have (inflexible) control over the iframe title via setting the `label` as they originally could.

gepifanio: @woowoowoo thats correct what Jared mentioned, do you prefer to be done differently? 

- About the `rel=0` I have added to the `video API` prop.

now you can send to the component:

image

`VideoAPI should be used for Youtube only`

woowoowoo: OK, sounds alright - I'm just unable to see it as a preview anywhere. Or am i missing something?

woowoowoo: In a perfect world, I don't think reusing the label is ideal.

The button label describes the action of a link, so "Play this video", whereas the title of the iframe should describe the content of the src document… "A youtube video fly-through of the Melbourne Connect building". They serve two different purposes, but for now, I guess it's fine.

jaredhenderson (approved):
The subheading font-family for the campaign banners may need to change if editors start using really long sentences. We can see how that goes, all looks fine otherwise

Issue #1684, opened 2021-07-02
[Feature] G2 footer-alt implemented in G2 sites


jaredhenderson:
### This is a placeholder ticket to block old footer issues
- [x] Update to footer-alt in FAC
- [x] Update to footer-alt in G2 CMS layouts

### When new footer is implemented
- [x] Old G2 footer tickets (a11y, bugs, etc.) can be closed
- [x] G2 old footer should be deprecated and removed from pattern-lib

PR #1694, opened 2021-07-13
chore(component): remove page footer component

gepifanio:
# Description

Fixes https://github.com/unimelb/pattern-lib/issues/1684

# Checklist:

- [x] Snapshots tested
- [x] A11y tested

# Reviewer Checklist

- [ ] A11y tested


components/print.css
```diff
@@ -10,8 +10,6 @@
   }
 
   .page-header,
```
jaredhenderson, line 12: Do you need to add some kind of class referencing the footer here, so the print styles still work for that area?

gepifanio: Fixed that

jaredhenderson (commented):
Looking at the old footer in print version versus new:
### Old
Screen Shot 2021-07-14 at 11 09 17 am

### New
Screen Shot 2021-07-14 at 11 07 19 am

- Looks like the blank section on the left needs to be removed from print display as well so the content can use that space

jaredhenderson:
One more thing - we are hiding some fairly important compliance information from printing now: 
Screen Shot 2021-07-14 at 11 36 22 am

Can these numbers be retained for print? The links on the left can remain hidden.

gepifanio:
@jaredhenderson does like this looks fine?

image


That's what I could do to tidy up a bit the layout

jaredhenderson (approved):
Looks good. The print CSS doesn't have to be perfect, just needs to remove background colours and retain the important info, so this looks fine

PR #1703, opened 2021-07-19
remove(font-weight): normal and replace existing to regular

gepifanio:


components/_vars.css
```diff
@@ -120,7 +120,6 @@
   --ff-lead: 'Noto Serif', 'Georgia', 'Times New Roman', serif;
   --fw-semibold: 600;
   --fw-bold: 700;
```
jaredhenderson, line 122: This looks fine now, would you mind ordering the vars by their weight, just to keep that easy to understand? eg;

```
--fw-light: 300;
--fw-regular: 400;
--fw-semibold: 600;
--fw-bold: 700;
```

jaredhenderson (approved):
Looks good, can one section just be re-organised slightly...

PR #1705, opened 2021-07-21
chore(component): improve mega menu accessibility

gepifanio:
# Description

image

Fixes https://app.zenhub.com/workspaces/global-design-system-issues-tracker-603c265651b123001c916450/issues/unimelb/pattern-lib/1654

# Checklist:

- [ ] Snapshots tested
- [ ] A11y tested

# Reviewer Checklist

- [ ] A11y tested


components/megamenu/MegaMenu.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div
+  <header
```
jaredhenderson, line 2: Shouldn't this be `nav`?

components/megamenu/MegaMenu.vue
```diff
@@ -26,7 +26,7 @@
           id="sitemapmenu"
           ref="rootmenu"
           :class="['megamenu', isShowTopMenu && 'megamenu__with-top-menu']"
-          aria-label="Site"
+          aria-label="Primary navigation"
```
jaredhenderson, line 29: - Can you rename this label to 'Local site navigation menu'

components/megamenu/MegaMenuTopNavigation.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <nav>
+  <nav aria-label="Secondary navigation">
```
jaredhenderson, line 2: - Can you rename this label to 'Other key sites'

components/megamenu/MegaMenu.vue
```diff
@@ -1,5 +1,5 @@
 <template>
```
jaredhenderson, line 1: - in Storybook, the top level `nav` for the menu doesn't have a label, only the ones inside do. Can you add one to the parent as well so every `nav` has a label. Label should be `Navigation menu`

jaredhenderson (changes requested):

jaredhenderson (approved):
Looks good 👍

Issue #1710, opened 2021-07-23
[Issues] Footer and page background colour (G2 + G3)


jaredhenderson:
### Previously
In Gen 2, a navy background colour is set on the `body` so that when a page is short and the footer doesn't reach the bottom of the viewport, it blends in seamless with the bottom of the footer.

### Currently
In Gen 3, background colours are not set, so the pages default to white. On short pages, this results in the following:

![Screen Shot 2021-07-23 at 13.32.02.png](https://images.zenhubusercontent.com/5acae0cd4b5806bc2bce3338/b02e38e5-19c9-4686-9afb-646a5b5d57c9)

This makes the design feel a bit clunky and unfinished. The chevron SVGs are also missing.

**Any solution should be replicated in both G2 and G3 for consistency.**

Example G3 page:
https://cms.unimelb.edu.au/myaccount/lookup
https://students.unimelb.edu.au/sortsite-test

Example G2 page:
https://study.unimelb.edu.au/sortsite-test

### Potential fixes to be investigated
**Warning: all of these options have potential side-effects**
- Add `background-color` on `body` to match colour on bottom of footer. Result:

![Screen Shot 2021-07-23 at 13.34.20.png](https://images.zenhubusercontent.com/5acae0cd4b5806bc2bce3338/621b5afa-a91e-42c5-8f40-8e37b9a4c26b)

  - Would need to ensure page content background is still white

- Add a `min-height` to the `main` content block so pages have a minimum height; suggest `25rem`
  - this is probably not ideal, it would be better if the page itself was always 100% browser height so this wasn't necessary, but that could be challenging

- Investigate and resolve missing chevron icons

gepifanio:
pr pattern-lib: https://github.com/unimelb/pattern-lib/pull/1712
pr gen3: https://github.com/unimelb/Gen3/pull/262
pr gen3 missing icons: https://github.com/unimelb/Gen3/pull/261

PR #1712, opened 2021-07-26
Chore: css body and main updates

gepifanio:
# Description

Fixes https://github.com/unimelb/pattern-lib/issues/1710

- Changed background-color on the body to match colour on the bottom of the footer.
- Added a min-height to the main content block so pages have a minimum height of 25rem
- That fixes the footer overflow, however as the page header colour is another navy you can see the page header overflow colour darker with this change:

image

# Checklist:

- [ ] Snapshots tested
- [ ] A11y tested

# Reviewer Checklist

- [ ] A11y tested


jaredhenderson:
- Can you provide more information please? URL example? Where are you seeing this issue? That 'overflow' should not be occurring in a header in the first place.

gepifanio:
> * Can you provide more information please? URL example? Where are you seeing this issue? That 'overflow' should not be occurring in a header in the first place.

That overflow happens only if you scroll backward when the header jumps down, is unlikely a user would be scrolling past the header, also when the user stops scrolling it would scroll back for the page header to stick at the top.

jaredhenderson:
> URL example? Where are you seeing this issue? 

What URL/environment/browser are you talking about? I can't do much with just a picture. Have you compared this to previous behaviour?

gepifanio:
> > URL example? Where are you seeing this issue?
> 
> What URL/environment/browser are you talking about? I can't do much with just a picture. Have you compared this to previous behaviour?

We don't have a preview link for that, The photo example is what I have tested in a live page changing the CSS properties, and Storybook cant replicate that. 

But basically, before the header colour and the page body colour were the same, you wouldn't see the colour difference whereas now they are different.

jaredhenderson (approved):
They weren't quite the same before - the header is Brand `navy-dark` and the page bg was `background-secondary-navy` which is only a little bit darker, so it was less obvious.

What browser is this happening in? I can't replicate it in Safari, Firefox or Chrome on desktop.

Anyway if that's all it is, we can't do much about it - we can't set the body colour to two different things. Probably looks worse with white so this is fine by me

PR #1713, opened 2021-07-26
enhancement(css): Use new grey-light colour to replace old background-primary-grey

jaredhenderson:
- Update `bg-alt` to use new `grey-light` so our older components match Today components
- Update other references to `background-primary-grey` to `grey-light`
- Disable `background-primary-grey` (it's still a brand colour so I haven't completely removed it)
- Remove custom `bg-grey-light` class as `bg-alt` covers this use case now


gepifanio (approved):
nice one!

PR #1736, opened 2021-08-17
feat: add in page nav from today into pattern-lib

gepifanio:
# Description

- Port over in-page navigation from study to pattern-lib

[preview](https://deploy-preview-1736--pattern-lib-unimelb.netlify.app/?selectedKind=Navigation&selectedStory=In%20page%20Navigation%20Today&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel)

Fixes https://jira.unimelb.edu.au/browse/TS-1295

## Type of change

# Checklist:

- [x] Snapshots tested
- [x] A11y tested

# Reviewer Checklist

- [ ] A11y tested


gepifanio:
preview mobile version icons doesn't have the gradient

preview:
image
image

localhost:
image

jaredhenderson (changes requested):
Looks pretty good. Can you make one tweak to the gradient background on the mobile nav buttons...
Currently the gradient has a hard edge because the starting position is `-23.89%`.

Screen Shot 2021-08-18 at 11 39 05 am


Can you adjust the backgrounds on both of these buttons to be more like the following:
`linear-gradient(90deg, rgba(237,242,245,0) 0%, rgba(237,242,245,.5) 40%, #edf2f5 95%)`

which looks like this:

Screen Shot 2021-08-18 at 11 38 45 am

jaredhenderson:
Re: use of `overlay` for `overflow` - non-standard property that isn't well supported:
https://caniuse.com/?search=overlay

jaredhenderson (approved):
Looks good 👍

Issue #1711, opened 2021-07-23
[Enhancement] G2 In-page navigation


jaredhenderson:
## Currently
The current **In-page navigation** component was built for FFAM and has a clunky and dated design.

![Screen Shot 2021-07-23 at 2.38.03 pm.png](https://images.zenhubusercontent.com/5acae0cd4b5806bc2bce3338/1595acdd-4058-4781-95c5-6fc5baa23fb4)
![Screen Shot 2021-07-23 at 2.38.18 pm.png](https://images.zenhubusercontent.com/5acae0cd4b5806bc2bce3338/3beb90a1-d527-4b58-8334-b52327245556)

[Pattern Library - In-page navigation](https://pattern-lib-unimelb.netlify.app/?selectedKind=Navigation&selectedStory=In%20page%20Navigation&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel)
[CMS Documentation](https://matrix-cms.unimelb.edu.au/web/mce/content-templates/in-page-navigation)

Example page:
https://study.unimelb.edu.au/dev-preview/open-day-2021/open-day-dev/open-day-about-dev/_nocache

Other examples:
https://matrix-cms.unimelb.edu.au/reports/content-templates-reports/in-page-nav

### [Figma designs >](https://www.figma.com/file/UF2630XlnsbudpZgvyZec0/Gen-2-and-3-Master-Components?node-id=2370%3A13712)

![Screen Shot 2021-07-23 at 2.40.08 pm.png](https://images.zenhubusercontent.com/5acae0cd4b5806bc2bce3338/846c4be8-8147-46da-a616-6f41c250c461)
![Screen Shot 2021-07-23 at 2.40.15 pm.png](https://images.zenhubusercontent.com/5acae0cd4b5806bc2bce3338/3ee5382b-d848-40ac-90a4-c40a07eee951)

The new style integrates the new design patterns from Today for the main layout and fixed navigation selector.

- Background colour should be fixed to this light grey so it stays consistent when it changes to fixed position mode
- These colours are consistent with the in-page navigation used on [Microcert pages](https://study.unimelb.edu.au/find/microcredentials/artificial-intelligence-and-women/)
![Screen Shot 2021-07-23 at 2.46.22 pm.png](https://images.zenhubusercontent.com/5acae0cd4b5806bc2bce3338/d0a42209-e9a8-4b8d-bd3a-c49d5dd08496)
[...]

PR #1737, opened 2021-08-18
chore: improve in page navigation component to use existent components

gepifanio:
# Description

- Removed `Dropdown` componet
- Tweaked existing `InPageNavigation` component to match new design
- In page navigation has new props and removed props
- Tweaked existing `SectionAlt` component
- Updated docs for `SectionAlt` and `InPageNavigation`

Fixes https://github.com/unimelb/pattern-lib/issues/1711

[preview](https://deploy-preview-1737--pattern-lib-unimelb.netlify.app/iframe.html?selectedKind=Navigation&selectedStory=In%20page%20Navigation)

## Type of change

# Checklist:

- [x] Snapshots tested
- [x] A11y tested

# Reviewer Checklist

- [ ] A11y tested


jaredhenderson:
Looking good. Some feedback:
- in Firefox, there's a weird dark line appearing during the transition on/off for the fixed navigation panel (doesn't appear in Chrome)

https://user-images.githubusercontent.com/34703139/130378129-2d576395-d777-4cb4-8626-057fcdf66003.mov

- Is it possible to remove the styling for the actual dropdown menu (screenshot below)? I'd rather just leave it to browser defaults

Screen Shot 2021-08-23 at 11 28 02 am

- Would it be possible to set the width of the menu to the width of the currently selected option? I know the select is sized based on the longest option by default... but if we could set it to the text width and add a transition for the `width` property, it might look a bit smoother

- Not sure if it is just this story, but when you resize the window some of the contents are causing something on the page to overflow and horizontal scrollbars appear - is this related to the component?

jaredhenderson:
Hi @gepifanio, thanks for adding the dynamic width. I think it won't be easy to rely on this as the font character width is not consistent, so we are seeing inconsistent widths from different option types.

To keep this item consistent, can you please make the select expand the width of the container? This will mean it takes up more space, but it also means it will behave the same each time someone encounters it.

jaredhenderson (changes requested):
Looking good from a functionality perspective, and the FFAM override is working well 👍 

Some comments around accessibility:
- [x] `aria-label` for select is hardcoded as 'Choose an option' which doesn't have any meaning. If it has to be hardcoded, this should be something like 'Table of contents options'
- [x] The related sections for these blocks should probably change from basic `div` elements to `nav` landmarks as that is their primary purpose

eg; the fixed position menu with select could be `nav` with `aria-label` 'Table of contents quick access menu' 

the normal links in the page section could be `nav` with `aria-label` 'Table of contents'

- [x] When I access the `select` by keyboard, there are two elements being focused in the same area - when I hit spacebar on the first one, it skips through each option. When I tab and hit spacebar on the next item, it opens the option menu and allows me to arrow up and down to choose any item

Should this just be one method?

jaredhenderson (approved):
✅ looks much better
✅ passes SortSite + SiteImprove (in Storybook)

nice work!

PR #1743, opened 2021-08-30
feat: port over app tabs

gepifanio:
# Description

[Preview](?selectedKind=Tabs%2FNew&selectedStory=AppTabs&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel)

Fixes # (issue number)

## Type of change

# Checklist:

- [ ] Snapshots tested
- [ ] A11y tested

# Reviewer Checklist

- [ ] A11y tested


jaredhenderson (commented):
Code etc. looks to be fine, I'm just not sure about the naming. `app` doesn't really mean anything in the pattern-lib context, and it doesn't describe the style of the tabs either.

Also complicating the issue is that the tabs have their own `alt` style as well.

Are we able to deprecate any older component to replace it with this, or are they too different?

PR #1769, opened 2021-09-09
fix: section alt bg alt spacing issue

gepifanio:
# Description

Section alt rules:

- [x] `bg-alt` + `bg-alt` = reduce padding top to 0
- [x] `bg-white` + `bg-white` =  reduce padding top to 0

[preview](https://deploy-preview-1769--pattern-lib-unimelb.netlify.app/?selectedKind=Navigation&selectedStory=In%20page%20Navigation%20Today&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel)

Figma: https://www.figma.com/file/E1vWS4bxzy21BI3De39DDn/UI-Kit?node-id=3493%3A1613

## Type of change

# Checklist:

- [x] Snapshots tested
- [x] A11y tested

# Reviewer Checklist

- [ ] A11y tested


gepifanio:
bg-white

image

jaredhenderson:
Looking good so far. There are a few places the rules don't cover (screenshots taken from story link):

Screen Shot 2021-09-22 at 1 38 30 pm
Screen Shot 2021-09-22 at 1 39 18 pm
Screen Shot 2021-09-22 at 1 40 32 pm

gepifanio:
> Looking good so far. There are a few places the rules don't cover (screenshots taken from story link):
> 
> Screen Shot 2021-09-22 at 1 38 30 pm
> 
> Screen Shot 2021-09-22 at 1 39 18 pm
> 
> Screen Shot 2021-09-22 at 1 40 32 pm

Yes I agree, but I haven't figured out how to target those inner bits with CSS, that's the best I could do, unless we do via JS

jaredhenderson:
We shouldn't be doing basic layout with JS. Did you try margins instead of padding, or a combination of both?

gepifanio:
> We shouldn't be doing basic layout with JS. Did you try margins instead of padding, or a combination of both?

Yes I tried using margins, but it wasn't working too

jaredhenderson:
I think I might have a solution for this. Not particularly fancy though, but it visually does the trick. I will need to re-design the example in Figma for your reference though, so I will have to come back to this once I have done my Study tickets for this sprint.

jaredhenderson:
Hey @gepifanio, as discussed, I think we can get the result we want with a slightly hacky method - instead of trying to reduce the `padding-bottom` on the `1st` section, we can easily target the `padding-top` on the `2nd` section instead.

So instead of `padding-bottom: 2.5rem` and `padding-top: 2.5rem` on the two sections, the first section can keep `padding-bottom: 5rem`, but the second section would have `padding-top: 0rem` instead.

Visually, this will achieve what we want, even though it might seem a bit odd when we inspect it with dev tools.

I have updated the [desktop example in Figma](https://www.figma.com/file/E1vWS4bxzy21BI3De39DDn/UI-Kit?node-id=3755%3A19882) to match this idea.

### Old concept
Sep-30-2021 16-06-52-before

### New concept
Sep-30-2021 16-09-00-after

PR #1771, opened 2021-09-10
fix tables p elements accessibility colour

gepifanio:
# Description

Fixes https://jira.unimelb.edu.au/browse/TS-1338

## Type of change

# Checklist:

- [x] Snapshots tested
- [x] A11y tested

# Reviewer Checklist

- [ ] A11y tested


components/tables/_tables.css
```diff
@@ -120,11 +120,19 @@
...
+      p {
+        color: rgb(var(--col-white));
+      }
     }
 
     &--white {
       background-color: rgb(var(--col-background-primary-white));
       color: rgb(var(--col-brand));
+
+      p {
```
jaredhenderson, line 133: Just use the standard `p` colour for other backgrounds (grey-mid-2 I think)

components/tables/_tables.css
```diff
@@ -135,6 +143,10 @@
       border-width: 2px;
       background-color: rgb(var(--col-blue-light));
       color: rgb(var(--col-brand));
+
+      p {
```
jaredhenderson, line 147: Just use the standard `p` colour for other backgrounds (grey-mid-2 I think)

jaredhenderson (commented):
Please modify the p colours to default (not brand) for the light bgs

jaredhenderson (commented):
One more change - the default colour for the white + highlight sections should also be grey, not brand (this must have been left over from when we used to have navy text for everything)

jaredhenderson (approved):

PR #1773, opened 2021-09-15
fix: table links accessibility colours

gepifanio:
Fixes https://jira.unimelb.edu.au/browse/TS-1337

[Preview](https://deploy-preview-1773--pattern-lib-unimelb.netlify.app/?selectedKind=Tables&selectedStory=Default&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel)

## Type of change

# Checklist:

- [x] Snapshots tested
- [x] A11y tested

# Reviewer Checklist

- [ ] A11y tested


components/tables/_tables.css
```diff
@@ -124,6 +124,14 @@
       p {
         color: rgb(var(--col-white));
       }
+
+      a {
+        color: rgb(var(--col-link-on-dark)) !important;
+
```
jaredhenderson, line 130: Do you need to specify any other states besides `visited`?

jaredhenderson (commented):
Can you check if hover/focus/whatever colours also need specifying here

components/tables/_tables.css
gepifanio: No, it would be following the default links states, we just need `visited` for that one :)

jaredhenderson: We don't want it to inherit the default link states since it's on an inverted background.

eg; it will inherit this `active` state colour - so it will be navy on navy again when active

Screen Shot 2021-09-15 at 2 44 21 pm

gepifanio: I am just following the existing structure for the links we have in the `_base.css` 
image


Do you reckon we should be changing that? 

I added a link in the storybook as example: [Preview](https://deploy-preview-1773--pattern-lib-unimelb.netlify.app/?selectedKind=Tables&selectedStory=Default&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel)

jaredhenderson: Where's the example? I can only see a link on white, which is not what we're trying to address

gepifanio: it should be in the table heading, probably wasn't ready when you tried to access the link, sorry about that.

jaredhenderson: I think it would be better to resolve this issue in `base` rather than duplicating the styles into `tables`

You just need to add `.table__row--inverted` (and whatever else) to the list of dark background styles for `p` and `a`

eg;
```
p {
  color: rgb(var(--col-grey-mid-2));

  .bg-inverted &,
  .bg-inverted-dark &,
  .bg-inverted-darker &,
  .bg-inverted--o95 &,
  .table__row--inverted {
    color: rgb(var(--col-white));
  }
}
```

gepifanio: good idea!

components/_base.css
```diff
@@ -165,9 +166,10 @@ a {
   .bg-inverted &,
   .bg-inverted-dark &,
   .bg-inverted-darker &,
-  .bg-inverted--o95 & {
+  .bg-inverted--o95 &,
+  .table__row--inverted & {
     &:not([class]) {
-      color: rgb(var(--col-link-on-dark));
+      color: rgb(var(--col-link-on-dark)) !important;
```
jaredhenderson, line 172: Are you able to remove the `!important`? It ends up overriding all the other link states, including `visited`...

jaredhenderson (approved):

PR #1786, opened 2021-09-27
fix element order for accessibility purposes

gepifanio:
# Description

- Fix HTML  structure so that the elements follow a logical order that makes sense when rendered without CSS.

Fixes https://jira.unimelb.edu.au/browse/TS-1504

## Type of change

# Checklist:

- [ ] Snapshots tested
- [ ] A11y tested
- [ ] Added proper labels to the PR and if there are any markup changes add `markup changes` label too.

# Reviewer Checklist

- [ ] A11y tested
- [ ] Crossbrowser tested
## Check List Before Merging


jaredhenderson (commented):
Do you know why so many `<!----->` seem to be appearing in rendered markup lately? I know it was having an impact on a 'Skip to content' link I saw in a report because the text was like `<a><!---->Skip to main content</a>` and the report was saying the text was empty, maybe because it looks like a comment?

Also is this linking to the correct Jira ticket? They seem unrelated

gepifanio:
> Do you know why so many `<!----->` seem to be appearing in rendered markup lately? I know it was having an impact on a 'Skip to content' link I saw in a report because the text was like `<a><!---->Skip to main content</a>` and the report was saying the text was empty, maybe because it looks like a comment?
> 
> Also is this linking to the correct Jira ticket? They seem unrelated

The code `<!----->` is from the vue template, basically where is using the `Select` but doesn't have the extra label on top because the vue has that `v-if` property to not render, it cause those empty blocks.

about the ticket it is to solve the ticket in this case, the issue says that label on top of the select using the position absolute is causing accessibility problems, and I changed the HTML structure order, to see if the tool would still be raising that as an issue because we still need to use the position absolute in that context, otherwise, we cant have the component matching the design.

jaredhenderson (approved):

PR #1792, opened 2021-10-01
chore: update new tabs styling to match new design requirements

gepifanio:
# Description

- update tabs to match new design
- add mobile scroll control

# Checklist:

- [ ] Snapshots tested
- [ ] A11y tested
- [ ] Added proper labels to the PR and if there are any markup changes add `markup changes` label too.

# Reviewer Checklist

- [ ] A11y tested
- [ ] Crossbrowser tested
## Check List Before Merging


jaredhenderson:
The scrolling nav is looking good. A couple of things:
- I've updated the [gradient spec in Figma](https://www.figma.com/file/E1vWS4bxzy21BI3De39DDn/UI-Kit?node-id=3913%3A19618) - there's a left and right one now with tweaked settings
- The button is overlapping the scroll bar like the Today one used to - I think in that case we changed the height to 80% to avoid it
Screen Shot 2021-10-07 at 12 17 52 pm

components/tabs/index.css
```diff
@@ -231,22 +247,26 @@
 
   &__tab {
     transition: background .15s ease;
-    text-align: center;
     flex: 1;
```
jaredhenderson, line 250: Sorry for the last minute request - can you remove `flex:1` so the tabs only take up their own space and stay on the left

Also can you adjust the padding on the left and right to be `1rem` so there's a bit more space...

PR #1793, opened 2021-10-03
feat: tweaks for new study area page

gepifanio:
# Description

#1790: chore: update filter widget / box and results 
#1791: chore: update page header alt

# Checklist:

- [x] Snapshots tested
- [x] A11y tested

# Reviewer Checklist

- [ ] A11y tested
- [ ] Crossbrowser tested
## Check List Before Merging


jaredhenderson:
Filter widget feedback:
- [ ] Please remove 'with secondary message' functionality
- [ ] Please remove 'with secondary message' story
- [ ] Can the story be updated to use the proper course cards for results?
- [ ] Filter area heading should only use `heading-sm` class for styling, not a custom class
- [ ] Can you remove the bold styling for the selects when they are modified - font weight should remain as regular

Header Alt feedback:
- [ ] White header looks fine with top alignment as long as tag is there; Dark header looks wrong as the text is too close to the top. Can we increase the `padding-top` on the container in the situation where a tag is not present? In the case where the tag is not present, the `padding-top` would need to be `6.375rem` for the heading to be in the same place as one with the tag (I've only checked desktop for this)

Should look like the below (roughly):
Screen Shot 2021-10-04 at 12 02 29 pm

- [ ] `min-height` needs to be lowered in the case where the padding is increased, otherwise the box gets larger
- [ ] The tag stripe is currently set to a custom `red` for Short Courses/Microcredentials. This colour should align with the course cards for those course types - should be `orange-mid-2`
- [ ] We need to be able to switch to a different colour for 'Study Area' (interest) - I believe this is currently `purple-dark`

jaredhenderson:
Looking good, some further tweaks for filters component:
- [ ] Please swap location of 'Apply' and 'Clear' buttons inside filter selects ('Apply' should be on top, matching the main filter display)
- [ ] Please add `heading-sm` class to the `h3` that shows 'No results' when there are no search results (this should be the only styling applied for that particular heading)
- [ ] Adjust border line colour at top of search results to match vertical border in Today alt section 
Screen Shot 2021-10-05 at 10 36 05 am

Header is looking good now.

I will also submit some updated designs for course cards to address [TS-1523 (lack of contrast)](https://jira.unimelb.edu.au/browse/TS-1523)

jaredhenderson (approved):
Looks good! I have closed the contrast issue relating to course cards as I don't believe the rules apply in this instance (as long as the text is OK) so you can ignore that now.

PR #1820, opened 2021-12-20
chore: chevron link styling and add chevron article list

gepifanio:
# Description

- Changed margin between button link text and chevron 
- Added button link text to article list

*Matrix CMS* attention

- Removed custom class `article-card__link`
- Markup changed component `CardArticleList`

[Preview Button Text Link](https://deploy-preview-1820--pattern-lib-unimelb.netlify.app/?selectedKind=Buttons%2FTypes&selectedStory=Text%20link&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel)
[Preview Card Article List](https://deploy-preview-1820--pattern-lib-unimelb.netlify.app/?selectedKind=Today%2FCards&selectedStory=Article%20List&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel)

Fixes https://jira.unimelb.edu.au/browse/TS-1679

# Checklist:

- [x] Snapshots tested
- [x] A11y tested

# Reviewer Checklist

- [ ] A11y tested
- [ ] Crossbrowser tested
## Check List Before Merging


jaredhenderson (changes requested):
Hey @gepifanio, this is mostly looking right but the chevron size isn't quite right for both cases.
- Text-only button chevron should equal 8 x 8px
- Chevron after article heading should be 10 x 10px

At the moment they are both using `.6rem` which is equating to 9.6px - I want to avoid using sub-pixel values whenever possible.

The difference between the two is small but relates to the text size the chevron is aligned to as [shown in Figma](https://www.figma.com/file/E1vWS4bxzy21BI3De39DDn/UI-Kit?node-id=4209%3A21479).

jaredhenderson:
Hey @gepifanio, the chevron sizes are better but I noticed the margin between the text and chevron looks bigger than expected for both the text-only button and the article heading links.

If you check them in the browser, they all have an extra `space` character at the end of the span which is making the gap appear too big.

Screen Shot 2021-12-21 at 10 48 31 am

Are you able to check if these can be removed?

gepifanio:
> 

Hey @jaredhenderson those extra spaces seem to be odd, I had a look and did some research.

I cant find how to get it fixed, so annoying 😞

jaredhenderson:
Hi @gepifanio, I'm going to log the whitespace bug as a separate issue. It appears to affect almost every Vue string I can see. There is actually a lot more whitespace in the DOM string than is being displayed, but that's only because the browser is stripping most of it out by default.

I've logged the details as [TS-1688](https://jira.unimelb.edu.au/browse/TS-1688) so that issue can be looked at separately as it appears to be a very undesirable bug.

jaredhenderson (approved):
Looks good. Hopefully the extra whitespace issue can be resolved separately

PR #1824, opened 2022-01-05
feat: section alt with video

gepifanio:
# Description

- Introduced section alt with video option
- new props `video` and `poster` in case the video haven't loaded
- Video background is purely for animation purposes and hidden from accessibility 
- Added new pause icon 
- Updated SectionAlt component and its documentation

[Preview](https://deploy-preview-1824--pattern-lib-unimelb.netlify.app/?selectedKind=Today%2FSection&selectedStory=Standard%20%28with%20Video%29&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel)

[Today miro board](https://miro.com/app/board/o9J_ln1yZW0=/)

Fixes https://jira.unimelb.edu.au/browse/TS-1695

## Type of change

# Checklist:

- [x] Snapshots tested
- [x] A11y tested
- [x] CSS utilizes BEM naming convention and structure

# Reviewer Checklist

- [ ] A11y tested
- [ ] Crossbrowser tested
## Check List Before Merging


jaredhenderson:
Hi @gepifanio, this is looking pretty good overall. A few things that weren't specced that should be adjusted:

- [ ] On page load, video should autoplay **and** loop without any user interaction, regardless of initial page location
- [ ] Video should continue to loop unless paused by the user
- [ ] If paused, video **should not be triggered by anything other than the play button**. No scrolling/viewport behaviour is needed - this is a simple background animation implementation, different to the 'portrait video with preview' behaviour
- [ ] Play/pause button should be a fixed width so it remains the same size when changing state (make it the size of the 'pause' state since that is the biggest)
- [ ] A backup image should always be available in case video is not supported (displayed behind video)
- [ ] There needs to be some kind of accessible description for the video for screenreaders, eg "Aerial footage of Melbourne University campus with Melbourne CBD and surrounds in background"
- [ ] There needs to be some kind of accessible description for the backup image for screenreaders
- The responsive states look good, I don't think designs are needed for this
- [ ] Check the play/pause button implementation - can buttons be focused by screenreaders easily? This is crucial as we are autoplaying video and must provide a way to stop it
- [ ] Play/pause button should have a 1px white border outline (on the top, left, right borders) so it can be seen when the video is very dark
- [ ] SVG on play/pause button doesn't seem to be accessible/have the right role or descriptions

gepifanio:
> Hi @gepifanio, this is looking pretty good overall. A few things that weren't specced that should be adjusted:
> 
> * [x]  On page load, video should autoplay **and** loop without any user interaction, regardless of initial page location
> * [x]  Video should continue to loop unless paused by the user
> * [x]  If paused, video **should not be triggered by anything other than the play button**. No scrolling/viewport behaviour is needed - this is a simple background animation implementation, different to the 'portrait video with preview' behaviour
> * [x]  Play/pause button should be a fixed width so it remains the same size when changing state (make it the size of the 'pause' state since that is the biggest)
> * [ ]  A backup image should always be available in case video is not supported (displayed behind video)
> * [ ]  There needs to be some kind of accessible description for the video for screenreaders, eg "Aerial footage of Melbourne University campus with Melbourne CBD and surrounds in background"
> * [ ]  There needs to be some kind of accessible description for the backup image for screenreaders
> * The responsive states look good, I don't think designs are needed for this
> * [ ]  Check the play/pause button implementation - can buttons be focused by screenreaders easily? This is crucial as we are autoplaying video and must provide a way to stop it
> * [x]  Play/pause button should have a 1px white border outline (on the top, left, right borders) so it can be seen when the video is very dark
> * [ ]  SVG on play/pause button doesn't seem to be accessible/have the right role or descriptions
> * [x]  Audio should be disabled (muted) for this component as it is only scoped to show a background animation. A user may accidentally upload a video with audio and we don't want the audio to ever play

Hey Jared,

I had a chat with Andy to clarify and help with some of the accessibility. 

He said this video is purely for animation purposes we should hide the video and the button from the accessibility tree.

The rest of the changes I have done ;)

jaredhenderson:
That's interesting that Andy suggested hiding everything, not what I would have expected.

There's still an issue with the SVG - I thought we didn't use `role=presentation` any more because it caused issues with SortSite, as it also doesn't have a label? Would be good to see SVG stuff documented for dev since it comes up a lot.

- [ ] check SVG implementation is definitely correct. are we documenting current SVG standards somewhere?
- [ ] play/pause button does not sit flush with the section top now - needs to increase by 1px due to border
- [ ] check SortSite results - I am getting two issues:
Screen Shot 2022-01-24 at 9 33 05 am
- [ ] Do we definitely not need to provide any alt-text for the video or poster image? Gonna be too late to change that later if we do need it and it's already in use

gepifanio:
I have sent SortSite an email to verify the rule:

image

As it seems to be a false positive.

PR #1830, opened 2022-02-09
chore(css):notices fix border and colors

jaredhenderson:
## Currently
In Gen 2, default paragraph colours are set on the `p` element. This makes paragraph text colours difficult to override with single classes, like `.bg-inverted`, unless you also specify the `p`.

This can have unpredictable side effects, for example:

1. Visit https://matrix-cms.unimelb.edu.au/study-fac/content/help
2. Change the notice type in the inverted section to `notice--warning`
3. See result:

Screen Shot 2022-02-09 at 3 49 13 pm

The `notice` class can be applied to `div` or `p` elements. Because this example is a `p`, it is inheriting the wrong text color because of the `.bg-inverted p` specificity.

**Additionally**, the notice component does not provide enough contrast when used on dark backgrounds (must be 3:1 or greater).

Relates to ticket [TS-1744](https://jira.unimelb.edu.au/browse/TS-1744)

## Proposed changes
**Notices:**
- add white `box-shadow` to notices to add a white "border" around them at all times. This will make them more visible on `alt` or `inverted` backgrounds

**Base:**
- move default font `color` from `html` element to `body` (like Gen3)
- don't set `p` default colour as it will inherit this from `body` anyway and it makes it harder to override
- tidy up `inverted` background CSS


gepifanio (approved):
Looks good!

